PR #10, opened 2019-06-28
Remove unsafeRun in .toResource, preserve interruptibility of Reservation in Resource. add .toResourceZIO

kaishh:
Breaks API in a binary-incompatible way by changing required implicits, although I think it's worth it to avoid deadlocking user's program


mijicd:
@kaishh you should have it now :)

kaishh:
Thanks!

interop-cats/jvm/src/main/scala/zio/interop/catszmanagedjvm.scala
```diff
@@ -69,13 +69,13 @@ final class CatsIOResourceSyntax[F[_], A](private val resource: Resource[F, A])
...
-    Resource.suspend(S.delay {
-      r.unsafeRun(managed.reserve) match {
-        case Reservation(acquire, release) =>
-          Resource.make(S.delay(r.unsafeRun(acquire)))(_ => S.delay(r.unsafeRun(release.unit)))
-      }
-    })
+  def toResourceZIO(implicit ev: Applicative[ZIO[R, E, ?]]): Resource[ZIO[R, E, ?], A] =
+    Resource
+      .make(managed.reserve)(_.release.unit)
+      .flatMap(Resource liftF _.acquire)
```
iravid: `evalMap` can be shorter here (same as what you wrote)

PR #28, opened 2019-07-31
Enable interop for scala.js

### LINKED ISSUE #20: Diverging implicit extension for UIO when importing cats instances
Scala will fail to compile code like this.
```scala
import cats.Monad
import cats.instances.either._
import scalaz.zio._
import scalaz.zio.interop.catz._

class ZIODiveringTest {
  Monad[UIO]
}
```

Removing the import for the either allows it to compile. Also works with option's instances. Does not happen for some reason when importing all instances.

Here is the error gotten
```
[error] D:\DevProjects\Stable\Ore\ore\app\ZIODiveringTest.scala:7:8: diverging implicit expansion for type cats.Monad[scalaz.zio.UIO]
[error] starting with method monoidZManagedInstances in trait CatsZManagedInstances
[error]   Monad[UIO]
[error]        ^
```

joroKr21:
Revive PR zio/zio#777
  * Test more thoroughly
  * Delete `Alternative` instance (unlawful)
  * Simplify some method implementations

There are two incompatibilities that need to be fixed (possibly in ZIO itself):
  * > bracket release is called on Completed or Error
  * > runCancelable is synchronous 

I don't think we can claim compatibility with cats-effect without fixing those.

**Edit**: I don't have the bandwidth to investigate the failures so I disabled the tests.
We can open separate issues for them.

This fixes #20


jdegoes:
> There are two incompatibilities that need to be fixed (possibly in ZIO itself):

Can you open two issues in ZIO for these, ideally with the tests restated using ZIO, rather than Cats Effect?

ghostdogpr:
Can you rebase to latest master? There was a renaming due to upgrading to the latest ZIO (in particular, `TaskR` => `RIO`)

joroKr21:
> Can you open two issues in ZIO for these, ideally with the tests restated using ZIO, rather than Cats Effect?

I'm not sure how to do that, the laws rely on classes from cats-effect. I will give it a try though. 

> Can you rebase to latest master? There was a lot of renaming due to upgrading to the latest ZIO (in particular, TaskR => RIO)

Rebased

neko-kai:
`runCancelable is synchronous` is just the lack of latch in upstream laws https://github.com/typelevel/cats-effect/pull/611

joroKr21:
@neko-kai nice!

neko-kai:
Seems like the same applies to `bracketReleaseIsCalledOnCompletedOrError` https://github.com/typelevel/cats-effect/pull/612

neko-kai:
I've also made a change to stop using `unsafeRunAsyncAndForget` since it logs failures (in addition to RTS doing that anyway and unlike cats IO implementations of the same)

neko-kai:
@joroKr21 
Is this PR good to merge now?

neko-kai:
```
[info] - Concurrent[Task].concurrent.uncancelable prevents Cancelled case *** FAILED ***
[info]   GeneratorDrivenPropertyCheckFailedException was thrown during property evaluation.
[info]    (Discipline.scala:12)
[info]     Falsified after 2 successful property evaluations.
[info]     Location: (Discipline.scala:12)
[info]     Occurred when passed generated values (
[info]       arg0 = <function1>,
[info]       arg1 = zio.ZIO$FlatMap@daeb5e4,
[info]       arg2 = zio.ZIO$FlatMap@2615c453
[info]     )
[info]     Label of failing property:
[info]       Expected: zio.ZIO$InterruptStatus@66e2b1e4
[info]   Received: zio.ZIO$InterruptStatus@70488c06
```

```
[info] - ApplicativeLocal[ZIO[Ctx, Error, ?]].applicativeLocal.local distributes over ap *** FAILED ***
[info]   GeneratorDrivenPropertyCheckFailedException was thrown during property evaluation.
[info]    (Discipline.scala:12)
[info]     Falsified after 1 successful property evaluations.
[info]     Location: (Discipline.scala:12)
[info]     Occurred when passed generated values (
[info]       arg0 = zio.ZIO$FlatMap@63f20abb,
[info]       arg1 = zio.ZIO$FlatMap@655029c5,
[info]       arg2 = <function1>
[info]     )
[info]     Label of failing property:
[info]       Expected: zio.ZIO$FlatMap@140e882f
[info]   Received: zio.ZIO$Read@434132d
```

Non-deterministic failures on CI? 🤔 
Or maybe these are also equalities broken by `genRace` and `genParallel` changing TestContext-based equality?

ghostdogpr:
I started to get weird diverging implicit issues after updating a project to cats 2.0.0-RC2:
```
diverging implicit expansion for type cats.Applicative[[+A]zio.ZIO[Any,Nothing,A]]
[error] starting with method catsKernelStdOrderForSortedSet in trait LowPrioritySortedSetInstancesBinCompat1
[error]   val test: UIO[Option[Unit]] = Option(ZIO.unit).sequence
```
I narrowed it down to `ioMonoidInstances `: by removing it, the diverging implicit error disappear. And I just saw that you removed it in this PR for other reasons but anyway that fixes the problem (tried with this PR branch).

joroKr21:
@neko-kai that's great work. Thank you so much for getting to the bottom of these tricky laws!

> I've also made a change to stop using unsafeRunAsyncAndForget since it logs failures (in addition to RTS doing that anyway and unlike cats IO implementations of the same)

Great, I hadn't thought of that :+1: 

> Non-deterministic failures on CI? thinking
Or maybe these are also equalities broken by genRace and genParallel changing TestContext-based equality?

Did you figure this out?

It's a bit crazy how much stack and heap we need, but it's not enough for scala.js
Is sbt not picking the `.jvmopts` up? Or something could be done in CircleCI config.

interop-cats/shared/src/test/scala/zio/interop/catzSpec.scala
```diff
@@ -0,0 +1,163 @@
...
+import zio.{ IO, _ }
+
+import scala.concurrent.Promise
+
+class catzSpec extends catzSpecBase {
+
+  def genUIO[A: Arbitrary]: Gen[UIO[A]] =
+    Gen.oneOf(genSuccess[Nothing, A], genIdentityTrans(genSuccess[Nothing, A]))
+
+  (1 to 10).foreach { i =>
```
joroKr21: If this is working now we can remove the loop

neko-kai:
@joroKr21 
> Did you figure this out?

Yeah, `io.zipPar(parIO)` in `genParallel` is non-deterministic wrt TestContext tasks, so I worked around it by replacing with `ZManaged(parIO).fork.use_(io)`...

OOM build failures on JS still remain though...

neko-kai:
> Is sbt not picking the .jvmopts up?

Yeah, sounds like they were ignored, moving options to .sbtopts worked

neko-kai:
Merged! 🎉  @joroKr21 Thank you!

Issue #24, opened 2019-02-08
Prototype cats effect 2.0


jdegoes:
Based on [this design](https://github.com/typelevel/cats-effect/issues/321). Could stick in the interop package and work out all the kinks before PR'ing to cats-effect.

Suggestion by @kaishh

ghostdogpr:
Closing this one, see https://github.com/zio/interop-cats/pull/36#issuecomment-530343216

PR #36, opened 2019-08-10
[WIP] Issue 24: Prototype cats effect 2.0

barambani:
This moves to `interop-cats` the changes in the original Zio's PR https://github.com/zio/zio/pull/828. The current structure is

Zio Hierarchy (1)

Ref https://github.com/zio/interop-cats/issues/24


kaishh:
👍

jdegoes:
@barambani @neko-kai Unfortunately this work is dead with Spiewak's CE 3.0 proposal. Although many things did make it in (not the bifunctor hierarchy, unfortunately).

I'm not sure what we do with this. Maybe it could be useful as an independent project, though if I were designing _just_ for ZIO users, it would look quite different (more simple / elegant).

@barambani I owe you a nice ribeye and a bottle of wine. Time to visit Dublin again!

jdegoes:
I will close this for now but we should think about whether to spawn this off into an independent project.

barambani:
@jdegoes, yes I noticed the CE 3 branch. Closing this makes sense. Also FWIW in the use cases I have I don't get any benefits from an hierarchy abstraction if is not able to support more than one effect type. I guess now the work on `interop-cats` will go on trying to get the support for the CE 3 if possible. So I will keep an eye here.
About the rest, :) no you don't owe me anything, actually thank you for all the learning and the great experience that was spending time on your projects. It was enlightening and inspiring. Yes, please come again, I will be very happy to meet you in Dublin again. The pleasure will be mine. Thanks 👍 .

PR #39, opened 2019-08-24
add sync instance for zmanaged

mschuwalow:


interop-cats/jvm/src/main/scala/zio/interop/catszmanagedjvm.scala
```diff
@@ -118,6 +120,12 @@ sealed trait CatsZManagedInstances1 {
 
 }
 
+sealed trait CatsZManagedInstances2 {
+
+  implicit def monadZManagedInstances[R, E]: Monad[ZManaged[R, E, ?]] = new CatsZManagedMonad
```
neko-kai: Doesn't MonadError already include Monad? There are no additional constraints on the MonadError instance, on reading it seems like Monad instance is redundant

mschuwalow: It was there before so I left it in and just changed the priority.
I'll throw it out. one sec

iravid:
The `bracketCase` implementation is a bit off. The release action can be interrupted.

neko-kai:
@iravid 
I wonder how does it pass cats-effect laws then 🤔

iravid:
Do the laws try to cancel during release?

neko-kai:
@iravid 
There's `acquireAndReleaseAreUncancelable` in BracketLaws, but we may not be generating IOs that could break the equality...

```scala
  def acquireAndReleaseAreUncancelable[A, B](fa: F[A], use: A => F[B], release: A => F[Unit]) =
    F.bracket(F.uncancelable(fa))(use)(a => F.uncancelable(release(a))) <-> F.bracket(fa)(use)(release)
```

The more interesting bracket-related laws are in Async* and up, those AFAIK do stress it properly

FWIW, cats.effect.Resource does not have a Sync instance...

iravid:
We'd need to use cancellation in the equality to see this happening. In any case, I can offer an implementation that behaves properly under cancellation tomorrow :-)

neko-kai:
@iravid Thanks 👍

mschuwalow:
@iravid @neko-kai Oh sry, I didn't notice. Implementing lawful instances is tricky O.o
@neko-kai thanks for making the changes / cleanup

mschuwalow:
I've taken another look at this and updated the implementation.
It should now be consistent with cats laws in terms of interruption.

One other change I've made is when the finalizer of release is run. I believe having it accumulate with the other finalizers of the managed is more consistent than running it eagerly.

On second thought I'm not sure whether it is a good idea to have this instance. While it is definitely useful in client code and I think can be consistent with the laws, the semantics feel a bit iffy.

I also wonder about tests. We could have tests for the more complex behavior of bracket either if we define async or concurrent instances, which I'm not quite sure we can, or by poking if with our own interruption, forking in tests.

@iravid @neko-kai what is your opinion on these?

interop-cats/shared/src/main/scala/zio/interop/catszmanaged.scala
```diff
@@ -104,43 +103,91 @@ trait CatsZManagedInstances extends CatsZManagedInstances1 {
...
+    with effect.Sync[ZManaged[R, Throwable, ?]] {
+
+  override final def bracketCase[A, B](acquire: zio.ZManaged[R, Throwable, A])(
+    use: A => zio.ZManaged[R, Throwable, B]
+  )(release: (A, ExitCase[Throwable]) => zio.ZManaged[R, Throwable, Unit]): zio.ZManaged[R, Throwable, B] =
+    ZManaged[R, Throwable, B] {
+      Ref.make[List[Exit[_, _] => ZIO[R, Nothing, Any]]](Nil).map { finalizers =>
+        Reservation(
+          acquire = ZIO.uninterruptibleMask { restore =>
+            for {
```
iravid: Great implementation @mschuwalow :-)

neko-kai:
> One other change I've made is when the finalizer of release is run. I believe having it accumulate with the other finalizers of the managed is more consistent than running it eagerly.

I wonder if this passing the laws could mean a problem with the `cats-effect` laws. As far as I understand, `release` MUST be run eagerly after `use` section is over, otherwise we'd easily have Bracket/Sync for all streaming types. 

/cc @alexandru IFF I'm not bothering you, could you share an opinion on that, since you know more about cats-effect laws?

mschuwalow:
> > One other change I've made is when the finalizer of release is run. I believe having it accumulate with the other finalizers of the managed is more consistent than running it eagerly.
> 
> I wonder if this passing the laws could mean a problem with the `cats-effect` laws.

That makes sense. I orginally did the change after looking at the laws and seeing equivalences such as: 
```scala
F.bracketCase(fa)(f)((_, _) => F.unit) <-> F.uncancelable(fa).flatMap(f)
```
which I believe should behave like this when directly translated to Zmanaged.

> As far as I understand, `release` MUST be run eagerly after `use` section is over, otherwise we'd
> easily have Bracket/Sync for all streaming types.

I also saw this mentioned in the ce-3 design discussion which cleans up the distinction up a bit for me:

> This is immensely difficult to encode with the current constructs. Higher level frameworks like fs2, Monix, and ZIO all have answers to this, but Cats Effect does not. Critically, the *reason* it is difficult to encode this use-case is the fact that `Resource` does not form a `Concurrent`, and thus lacks a `race` function. Unfortunately, the whole reason that `Resource` doesn't form a `Concurrent` comes down to the fact that `Concurrent` (in CE1) implies `Bracket`, and `Resource` cannot form a lawful `Bracket` since it is a dynamic region, not a static one.

This does point to there actually being no way to lawfully implement bracket as flatMap fundamentally behaves in a "wrong" way for this, if I understand correctly?

alexandru:
Hi @neko-kai,

Not sure what the issue is here and unfortunately I'm super busy these days, but ...

> One other change I've made is when the finalizer of release is run. I believe having it accumulate with the other finalizers of the managed is more consistent than running it eagerly.
>
> I wonder if this passing the laws could mean a problem with the cats-effect laws. As far as I understand, release MUST be run eagerly after use section is over, otherwise we'd easily have Bracket/Sync for all streaming types.

Note that yes, the intent for the `release` is for it to be executed immediately after `use`.

I don't think we have a law for it and when introducing laws it's good if we have very good reasons for it. And so I'm open to hearing opinions to why the `release` shouldn't be executed immediately after `use`.

Note that with `Bracket` the current expected (assumed) behavior does not prevent streaming data types from implementing `Bracket`. In fact `Observable` from Monix implements `Bracket` just fine. Basically if a stream emits `A` resources, then `release(A)` is called for each individual event, one after another.

The one problem that `Bracket` has for streaming data types is that the`acquire` in `bracket`  is uncancelable (uninterruptible) as a requirement, which currently can't work for data types such as `Iterant` or `fs2.Stream` — this is one place where we've got an `IO`-related assumption that leaked into `Bracket` without intention, see issue https://github.com/typelevel/cats-effect/issues/382

neko-kai:
@mschuwalow 
Would you be open to changing `bracket` such that it executes `release` eagerly? Which I believe is what Alex meant by:
> In fact Observable from Monix implements Bracket just fine. Basically if a stream emits A resources, then release(A) is called for each individual event, one after another.

Then I guess we can merge. I think the semantics in that case aren't _super_ strange, and there's a precedent in [fs2.Pull](https://github.com/functional-streams-for-scala/fs2/blob/829279df39c6cb779e512061ed6b4cdc8f622bb1/core/shared/src/main/scala/fs2/Pull.scala#L183) Bracket instance where the Pull has "delimited" bracketCase, while at the same time Pull `=` Stream where Stream has a separate region-based bracket.

mschuwalow:
Sure, I'll try to get this updated sometime this week

mschuwalow:
> @mschuwalow
> Would you be open to changing `bracket` such that it executes `release` eagerly? Which I believe is what Alex meant by:
> 
> > In fact Observable from Monix implements Bracket just fine. Basically if a stream emits A resources, then release(A) is called for each individual event, one after another.
> 
> Then I guess we can merge. I think the semantics in that case aren't _super_ strange, and there's a precedent in [fs2.Pull](https://github.com/functional-streams-for-scala/fs2/blob/829279df39c6cb779e512061ed6b4cdc8f622bb1/core/shared/src/main/scala/fs2/Pull.scala#L183) Bracket instance where the Pull has "delimited" bracketCase, while at the same time Pull `=` Stream where Stream has a separate region-based bracket.

I've updated the pr. Can you take another look?
I opted to implement this in a low level way insted of relying on bracket so that I don't have two swallow the typed error in release

interop-cats/shared/src/main/scala/zio/interop/catszmanaged.scala
```diff
@@ -160,3 +163,36 @@ private class CatsZManagedSemigroupK[R, E] extends SemigroupK[ZManaged[R, E, ?]]
...
+  )(
+    release: (A, cats.effect.ExitCase[Throwable]) => zio.ZManaged[R, Throwable, Unit]
+  ): zio.ZManaged[R, Throwable, B] =
+    ZManaged.fromEffect {
+      ZIO.uninterruptibleMask { restore =>
+        for {
+          res <- acquire.reserve
+          fin = res.release
+          a   <- res.acquire.onError(c => fin(Exit.halt(c)))
+          b <- restore(use(a).use(ZIO.succeed)).run.flatMap { exit =>
```
neko-kai: This means the returned `B` is released early, so it's "invalid"/deallocated outside of bracket if it had a finalizer.
The usecase i have for this i is to express "allocate temporary resources for allocation of B", for this B must remain allocated outside of .use. Do you think it would make sense here to run reserve and acquire here, but make the finalizer Managed? (Same as ZManaged.preallocate)

mschuwalow: Ah I see, that makes sense as well.
I was thinking that typically `B` would have a dependency on `A`, and would be defunct after the release / `A`s finalizer is run. Also wouldn't that imply that we delay running release immediately?

So you would propose something like this (ignoring resource safety / interruption)?
```
    ZManaged {
      Ref.make[List[Exit[_, _] => URIO[R, _]]](Nil).map { finalizers =>
        Reservation(
          for {
            resA <- acquire.reserve
            a    <- resA.acquire
            resB <- use(a).reserve
            b    <- resB.acquire
            _    <- release(a, exitToExitCase(Exit.unit)).use_(ZIO.unit) // when exactly should this be run?
            _    <- resA.release(Exit.unit)
            _    <- finalizers.update(resB.release :: _)
          } yield b,
          exitU =>
            for {
              fs    <- finalizers.get
              exits <- ZIO.foreach(fs)(_(exitU).run)
            } yield Exit.collectAllPar(exits)
        )
      }
    }
```

neko-kai: @mschuwalow 
I think it makes sense to run all actions in bracket order (run release __action__ eagerly after __use__ action), but defer ALL finalizers as is normal in ZManaged control flow, e.g.:

```scala
ZManaged.unwrap {
  ZIO.uninterruptibleMask { restore =>
    for {
      resA <- acquire.reserve
      finA = resA.release
      a    <- resA.acquire.onError(c => finA(Exit.halt(c)))

      resB  <- restore(use(a).reserve.uninterruptible).onError(c => finA(Exit.halt(c)))
      finB  = resB.release
      exitB <- restore(resB.acquire).run.onError(c => finB(Exit.halt(c)).ensuring(finA(Exit.halt(c))))

      allocC <- release(a, exitToExitCase(exitB)).preallocate.onError { c =>
                 finB(Exit.halt(c)).ensuring(finA(Exit.halt(c)))
               }
      finC = (e: Exit[Any, Any]) => allocC.reserve.map(_.release(e))
      b <- ZIO.done(exitB).onError { c =>
            finC(Exit.halt(c)).ensuring(finB(Exit.halt(c)).ensuring(finA(Exit.halt(c))))
          }
    } yield ZManaged.reserve(Reservation(ZIO.succeed(b), e => finC(e).ensuring(finB(e).ensuring(finA(e)))))
  }
}
```

Since you used force push, I'm not sure if this isn't equivalent to what you had before, but this seems similar to your initial version using ZManaged.foldCauseM – I now believe this makes the most sense, it both adheres to running `release` eagerly after `use`, but also the values aren't invalidated ahead of time by running their finalizers (if zmanaged finalizers finalize instead of the release action)

mschuwalow: I'm on board with this! I'll update the pr in a sec

mschuwalow: @neko-kai 
I've updated the pr. Can you check whether this is what you had in mind? I believe this should be equivalent to the snippet you posted.

interop-cats/shared/src/main/scala/zio/interop/catszmanaged.scala
```diff
@@ -160,3 +163,51 @@ private class CatsZManagedSemigroupK[R, E] extends SemigroupK[ZManaged[R, E, ?]]
...
+        Reservation(
+          ZIO.uninterruptibleMask { restore =>
+            for {
+              a <- for {
+                    resA <- acquire.reserve
+                    _    <- finalizers.update(resA.release :: _)
+                    a    <- restore(resA.acquire)
+                  } yield a
+              exitB <- (for {
+                        resB <- use(a).reserve
```
neko-kai: wrapping it in `restore(_.uninterruptible)` would allow `use` to be interrupted before starting, if the fiber was interrupted during `acquire` – behavior closer to ordinary bracket, where it's legal for `use` to be interrupted immediately before starting if the fiber is interrupted in acquire.

mschuwalow: ah cool, I wasn't aware of that

neko-kai:
Thank you and sorry for the trouble! :pray:

neko-kai:
Thank you and sorry for the trouble! :pray:

PR #48, opened 2019-08-29
Trying to fix cats-effect laws wrt TestContext (currently TestContext is bypassed)

neko-kai:


joroKr21:
Hmm, what's wrong with the tests?
Everything is green for me :thinking:

neko-kai:
@joroKr21
They work right now only bcs of workarounds in genRace/genParallel IIRC https://github.com/typelevel/cats-effect/pull/612#issuecomment-526554265

neko-kai:
Fixed in https://github.com/zio/interop-cats/pull/156

PR #67, opened 2019-09-21
Add ZStream instances

regiskuckaertz:
Pending https://github.com/zio/zio/pull/1725


neko-kai:
There are some compile failures...

```
[error] /home/circleci/project/interop-cats/shared/src/main/scala/zio/stream/cats.scala:60:79: value either is not a member of zio.stream.ZStream[R,E,A]
[error]   override def attempt[A](fa: ZStream[R, E, A]): ZIO[R, E, Either[E, A]] = fa.either
[error]                                                                               ^
[error] /home/circleci/project/interop-cats/shared/src/main/scala/zio/stream/cats.scala:72:82: value as is not a member of zio.stream.ZStream[R,E,A]
[error] did you mean map or tap?
[error]   override final def as[A, B](fa: ZStream[R, E, A], b: B): ZStream[R, E, B] = fa.as(b)
[error]                                                                                  ^
[error] /home/circleci/project/interop-cats/shared/src/main/scala/zio/stream/cats.scala:79:58: value unit is not a member of object zio.stream.ZStream
[error] did you mean wait?
[error]   override final def unit: ZStream[R, E, Unit] = ZStream.unit
[error]                                                          ^
[error] /home/circleci/project/interop-cats/shared/src/main/scala/zio/stream/cats.scala:94:9: value bimap is not a member of zio.stream.ZStream[R,A,B]
[error] did you mean map?
[error]     fab.bimap(f, g)
[error]         ^
```

regiskuckaertz:
@neko-kai yes, these are combinators I added yesterday, pending next milestone or ... the one dot oh 😄

interop-cats/shared/src/main/scala/zio/stream/cats.scala
```diff
@@ -0,0 +1,122 @@
...
+
+private class CatsMonad[R, E]
+    extends CatsApplicative[R, E]
+    with Monad[ZStream[R, E, *]]
+    with StackSafeMonad[ZStream[R, E, *]] {
+  override final def flatMap[A, B](fa: ZStream[R, E, A])(f: A => ZStream[R, E, B]): ZStream[R, E, B] = fa.flatMap(f)
+
+  override final def widen[A, B >: A](fa: ZStream[R, E, A]): ZStream[R, E, B] = fa
+  override final def map2[A, B, Z](fa: ZStream[R, E, A], fb: ZStream[R, E, B])(f: (A, B) => Z): ZStream[R, E, Z] =
+    fa.zipWith(fb) { case (Some(a), Some(b)) => Some(f(a, b)); case _ => None }
```
iravid: Pretty sure there's a law that says `map2` (and `ap`) need to be consistent with `flatMap`. This function applies the function pointwise on the streams, but to be consistent with `flatMap` we need the cartesian product. That's `crossWith`.

interop-cats/shared/src/main/scala/zio/stream/cats.scala
```diff
@@ -0,0 +1,122 @@
...
+  override final def as[A, B](fa: ZStream[R, E, A], b: B): ZStream[R, E, B] = fa.as(b)
+}
+
+private trait CatsApplicative[R, E] extends Applicative[ZStream[R, E, *]] {
+  override final def pure[A](a: A): ZStream[R, E, A]                              = ZStream.succeed(a)
+  override final def map[A, B](fa: ZStream[R, E, A])(f: A => B): ZStream[R, E, B] = fa.map(f)
+
+  override final def unit: ZStream[R, E, Unit] = ZStream.unit
+  override final def whenA[A](cond: Boolean)(f: => ZStream[R, E, A]): ZStream[R, E, Unit] =
+    ZStream.fromEffect(ZIO.effectTotal(f).when(cond))
```
iravid: isn't this `if (cond) f.unit else ZStream.fromEffect(UIO.unit)`? It looks like `f` won't actually be evaluated.

interop-cats/shared/src/main/scala/zio/stream/cats.scala
regiskuckaertz: Got it 👍

interop-cats/shared/src/main/scala/zio/stream/cats.scala
regiskuckaertz: It will be if `cond` is true 🤔

iravid: `a = ZIO.effectTotal(f): ZIO[R, E, ZStream[R, E, A]]`
`b = a.when(cond): ZIO[R, E, Unit]` <== `f` is discarded here and not drained/evaluated
`c = ZStream.fromEffect(b): ZStream[R, E, Unit]` <== draining `c` will be a no-op

regiskuckaertz:
So, there is an `interop-monix` and I'm wondering whether it wouldn't make more sense to create `interop-fs2` rather than add conversions in here. I imagine we would want conversions to/from `Observable`/`Iterant` as well and those would naturally go in `interop-monix`. wdyt?

iravid:
@regiskuckaertz I don't mind that. Let's ask @neko-kai though

regiskuckaertz:
@neko-kai if it can wait until tomorrow, I’ll add the laws tests so this can get merged if it looks good to you

neko-kai:
@regiskuckaertz Sure! 👍

regiskuckaertz:
@neko-kai @iravid I was a bit late, sorry. Can you please take a look and let me know if there's anything missing?

regiskuckaertz:
oh, yeah let me have a look.

PR #83, opened 2019-10-06
Add Parallel instance for pointwise zipping of streams

regiskuckaertz:


regiskuckaertz:
🤔 where is that `Arbitrary[Z]` being summoned

regiskuckaertz:
Fixed, if you don't mind approving again 😅

PR #104, opened 2019-11-02
Add an ArrowChoice instances for RIO and URIO

joroKr21:
Due to partial-unification `ZIO[*, E, *]` is not an inferrable type.
Which means that syntax doesn't work with the more general instance.
However it does work if we use `RIO`, e.g.

```scala
def foo(rio: RIO[String, Int]): RIO[Int, String] =
  rio.dimap(_.toString)(_.toString)
```

The same `foo` would not work with `ZIO`.


joroKr21:
Any opinions whether / how we should test this?

iravid:
@joroKr21 Latest ZIO Test has an `assertCompiles` assertion that you could use to test the inference

joroKr21:
Hmm, looks like it's not released yet.
On the other hand to assert that something compiles we can just write the code :smile: 
Ok I think I will add a very minimal test for now let me know if we need more.

joroKr21:
Okay, I think it would also be good to add an instance for `URIO`.

interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -101,9 +101,16 @@ abstract class CatsInstances extends CatsInstances1 {
...
+    new CatsBifunctor[R]
 
-  implicit def zioArrowInstance[E]: ArrowChoice[ZIO[*, E, *]] = new CatsArrow[E]
+  implicit val rioArrowInstance: ArrowChoice[RIO] =
+    new CatsArrow[Throwable]
+
+  implicit val urioArrowInstance: ArrowChoice[URIO] =
+    new CatsArrow[Nothing]
+
+  implicit def zioArrowInstance[E]: ArrowChoice[ZIO[*, E, *]] =
```
neko-kai: @joroKr21 
I'd pull this down into a low priority trait because the choice for `ZIO[*, Throwable, *]` even if not ambiguous now due to specifics, is ambiguous in actuality and may become ambiguous if implicits are improved in the future.

joroKr21: Sure, will do. 

Note however that I wouldn't say more ambiguity would be an improvement. If I wrote an implicit for some concrete type by hand, I want it to be preferred over the generic one.

adamgfraser:
@joroKr21 You can negate an assertion to assert that code does not compile. I know it isn't in the latest release yet but when the next release comes out it seems like `assertCompiles` could be a good solution for you.

interop-cats/shared/src/main/scala/zio/interop/cats.scala
neko-kai: @joroKr21 
Hmm, seems like dotty/scala already behaves like that https://scastie.scala-lang.org/Xa5gdVA5RDOKTEyF5Bob3A https://scastie.scala-lang.org/LoIvb5yiR42lbyPCZMSdLA
So no reason for LP trait here

joroKr21:
> @joroKr21 is it ok to merge as is?

Yes, I think it can be merged.

PR #111, opened 2019-11-17
Add ZIO/RIO/IO/Task.concurrentEffect & concurrentEffectWith helpers

neko-kai:
UIO/URIO are omitted because they can't host `Throwable`


interop-cats/shared/src/main/scala/zio/interop/CatsConcurrentEffectSyntax.scala
```diff
@@ -0,0 +1,33 @@
+package zio.interop
+
+import cats.effect.ConcurrentEffect
+import zio.{ IO, RIO, Runtime, Task, ZIO }
+
+import scala.language.implicitConversions
+
+trait CatsConcurrentEffectSyntax {
+  implicit final def ZIOConcurrentEffectOps(@deprecated("", "") zio: ZIO.type): CatsConcurrentEffectSyntax.zioOps.type =
```
ghostdogpr: Why deprecated?

neko-kai: @ghostdogpr to silence fatal warning that parameter is unused

PR #145, opened 2020-02-12
Update to latest zio

ioleo:
I'm experimenting with new DI encoding (Has & ZLayer) from the latest ZIO and I needed to update interop-cats to play with http4s/doobie.
This ofcourse will need to be updated when the next version lands (be it `1.0.0` or `1.0.0-RC18`).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -47,19 +47,19 @@ lazy val interopCats = crossProject(JSPlatform, JVMPlatform)
   .settings(buildInfoSettings)
   .settings(
     libraryDependencies ++= Seq(
-      "dev.zio"       %%% "zio"                  % "1.0.0-RC17",
-      "dev.zio"       %%% "zio-streams"          % "1.0.0-RC17" % Optional,
-      "dev.zio"       %%% "zio-test"             % "1.0.0-RC17" % Optional,
-      "org.typelevel" %%% "cats-effect"          % "2.0.0" % Optional,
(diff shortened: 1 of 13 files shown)


ioleo:
@pierangeloc After all day playing around I come to discord to discover you did the same work today :joy:

pierangeloc:
@ioleo my branch https://github.com/pierangeloc/interop-cats/tree/use-zio-rc18-snapshot
 works for 2.13 (including tests) but fails for other Scala versions. I managed to `publishLocal` and use it with http4s (no doobie yet though).
I am also writing some documentation on the new ZLayer approach, to eventually contribute it to ZIO website. Just to avoid double work :-), are you busy doing the same?

mijicd:
I don't think we should depend on snapshots. It's super convenient for development for sure, but it feels a bit odd to have it here.

ghostdogpr:
@mijicd I think it's good to have that PR ready although we don't merge it until the actual RC is released. Several times in the past we released zio and everybody was expecting interop-cats to follow but there was substantial work to do. It's much better if that work is done beforehand.

mijicd:
@ghostdogpr with this I agree 100% :)

ioleo:
@pierangeloc I'm playing around for myself, but a thought to create some docs/blog post/whatever crossed my mind. I'm quite busy lately so it might take a while, by that time I will first check if you had not posted something already :)

Issue #22, opened 2019-04-24
Cats Effect: Queue


jdegoes:
Introduce an effect-polymorphic class atop `ZQueue`, with a default implementation for any Cats Effect data type (Cats IO, Monix Task, etc.).

```scala
package scalaz.zio.interop

trait Queue[F[_], +A] {
  ...
}
object Queue {
  ...
}
```

The methods will all delegate to ZIO `ZQueue`, but they will take the `ZIO` effect returned by `ZQueue` methods and lift it into any effect `F[_]` that is suitable equipped.

calvinlfer:
It’s been quite a while on this one. @adamgfraser helped me make a lot of progress so I’d like to continue working on this if that’s alright

calvinlfer:
Completed by https://github.com/zio/interop-cats/pull/146

PR #146, opened 2020-02-17
Allow ZIO's Queue to be used by Cats Effect

calvinlfer:
Took a shot at implementing https://github.com/zio/interop-cats/issues/22 with @adamgfraser's help. Would love some feedback so we can get this in 😸

### DIFF
--- a/interop-cats/shared/src/main/scala/zio/interop/Queue.scala
+++ b/interop-cats/shared/src/main/scala/zio/interop/Queue.scala
@@ -22,27 +22,27 @@ import zio.{ Queue => ZioQueue, Runtime, UIO, ZEnv }
 object Queue {
 
   /**
-   * @see [[ZioQueue.bounded]]
+   * @see ZioQueue.bounded
    */
-  final def bounded[F[+_], R, A](capacity: Int)(implicit R: Runtime[ZEnv], F: LiftIO[F]): F[Queue[F, A]] =
+  final def bounded[F[+_], A](capacity: Int)(implicit R: Runtime[ZEnv], F: LiftIO[F]): F[Queue[F, A]] =
     create(ZioQueue.bounded[A](capacity))
 
   /**
-   * @see [[ZioQueue.dropping]]
+   * @see ZioQueue.dropping
    */
-  final def dropping[F[+_], R, A](capacity: Int)(implicit R: Runtime[ZEnv], F: LiftIO[F]): F[Queue[F, A]] =
+  final def dropping[F[+_], A](capacity: Int)(implicit R: Runtime[ZEnv], F: LiftIO[F]): F[Queue[F, A]] =
     create(ZioQueue.dropping[A](capacity))
 
   /**
-   * @see [[ZioQueue.sliding]]
+   * @see ZioQueue.sliding
    */
-  final def sliding[F[+_], R, A](capacity: Int)(implicit R: Runtime[ZEnv], F: LiftIO[F]): F[Queue[F, A]] =
+  final def sliding[F[+_], A](capacity: Int)(implicit R: Runtime[ZEnv], F: LiftIO[F]): F[Queue[F, A]] =
     create(ZioQueue.sliding[A](capacity))
 
   /**
-   * @see [[ZioQueue.unbounded]]
+   * @see ZioQueue.unbounded
    */
-  final def unbounded[F[+_], R, A](implicit R: Runtime[ZEnv], F: LiftIO[F]): F[Queue[F, A]] =
+  final def unbounded[F[+_], A](implicit R: Runtime[ZEnv], F: LiftIO[F]): F[Queue[F, A]] =
     create(ZioQueue.unbounded[A])
 
   private final def create[F[+_], A](in: UIO[ZioQueue[A]])(implicit R: Runtime[ZEnv], F: LiftIO[F]): F[Queue[F, A]] =
--- a/interop-cats/shared/src/main/scala/zio/interop/Queue.scala
+++ b/interop-cats/shared/src/main/scala/zio/interop/Queue.scala
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2017-2019 John A. De Goes and the ZIO Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio.interop
+
+import cats.effect.LiftIO
+import zio.{ Queue => ZioQueue, Runtime, UIO, ZEnv }
+
+object Queue {
+
+  /**
+   * @see [[ZioQueue.bounded]]
+   */
+  final def bounded[F[+_], R, A](capacity: Int)(implicit R: Runtime[ZEnv], F: LiftIO[F]): F[Queue[F, A]] =
+    create(ZioQueue.bounded[A](capacity))
+
+  /**
+   * @see [[ZioQueue.dropping]]
+   */
+  final def dropping[F[+_], R, A](capacity: Int)(implicit R: Runtime[ZEnv], F: LiftIO[F]): F[Queue[F, A]] =
+    create(ZioQueue.dropping[A](capacity))
+
+  /**
+   * @see [[ZioQueue.sliding]]
+   */
+  final def sliding[F[+_], R, A](capacity: Int)(implicit R: Runtime[ZEnv], F: LiftIO[F]): F[Queue[F, A]] =
+    create(ZioQueue.sliding[A](capacity))
+
+  /**
+   * @see [[ZioQueue.unbounded]]
+   */
+  final def unbounded[F[+_], R, A](implicit R: Runtime[ZEnv], F: LiftIO[F]): F[Queue[F, A]] =
+    create(ZioQueue.unbounded[A])
+
+  private final def create[F[+_], A](in: UIO[ZioQueue[A]])(implicit R: Runtime[ZEnv], F: LiftIO[F]): F[Queue[F, A]] =
+    toEffect(in.map(new CQueue[F, A, A](_)))
+}
--- a/interop-cats/shared/src/test/scala/zio/interop/catzQueueSpec.scala
+++ b/interop-cats/shared/src/test/scala/zio/interop/catzQueueSpec.scala
@@ -14,7 +14,7 @@ object catzQueueSpec
     extends DefaultRunnableSpec({
       def boundedQueueTest[F[+_]](implicit F: Concurrent[F], R: Runtime[ZEnv]): F[TestResult] =
         for {
-          q  <- Queue.bounded[F, Any, Int](1)
+          q  <- Queue.bounded[F, Int](1)
           _  <- q.offer(1)
           f  <- q.offer(2).start
           r1 <- q.takeAll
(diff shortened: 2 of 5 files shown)


interop-cats/shared/src/main/scala/zio/interop/queue/CQueue.scala
```diff
@@ -0,0 +1,72 @@
+package zio.interop.queue
```
adamgfraser: I think this can just be in the `cats.interop` package like the `Schedule` compatibility implementation to mirror the structure in ZIO.

interop-cats/shared/src/main/scala/zio/interop/queue/CQueue.scala
```diff
@@ -0,0 +1,72 @@
...
+  def poll(implicit R: Runtime[Any], F: LiftIO[F]): F[Option[B]] = toEffect(q.poll)
+}
+
+object CQueue {
+  import zio.interop.catz._
+  import zio.Runtime
+
+  private type CompatQueue[-A, +B] = zio.ZQueue[Any, Throwable, Any, Throwable, A, B]
+
+  private[queue] def toEffect[F[+_], R, A](zio: RIO[R, A])(implicit R: Runtime[R], F: LiftIO[F]): F[A] =
```
adamgfraser: These definitions are duplicated in the `Schedule` compatibility implementation. Why don't we extract them to the package object and make them package private?

interop-cats/shared/src/main/scala/zio/interop/queue/CQueue.scala
```diff
@@ -0,0 +1,72 @@
...
+  def mapM[C](f: B => F[C])(implicit R: Runtime[Any], E: Effect[F]) = new CQueue(q.mapM(b => fromEffect(f(b))))
+
+  def poll(implicit R: Runtime[Any], F: LiftIO[F]): F[Option[B]] = toEffect(q.poll)
+}
+
+object CQueue {
+  import zio.interop.catz._
+  import zio.Runtime
+
+  private type CompatQueue[-A, +B] = zio.ZQueue[Any, Throwable, Any, Throwable, A, B]
```
adamgfraser: We are only using this type alias one place so we can probably just inline it.

interop-cats/shared/src/main/scala/zio/interop/queue/CQueue.scala
```diff
@@ -0,0 +1,72 @@
...
+    new CQueue(q.contramapM(c => fromEffect(f(c))))
+
+  def filterInput[A0 <: A](f: A0 => Boolean): CQueue[F, A0, B] = new CQueue(q.filterInput(f))
+
+  def filterInputM[A0 <: A](f: A0 => F[Boolean])(implicit R: Runtime[Any], E: Effect[F]) =
+    new CQueue(q.filterInputM((a0: A0) => fromEffect(f(a0))))
+
+  def map[C](f: B => C): CQueue[F, A, C] = new CQueue(q.map(f))
+
+  def mapM[C](f: B => F[C])(implicit R: Runtime[Any], E: Effect[F]) = new CQueue(q.mapM(b => fromEffect(f(b))))
```
adamgfraser: Would suggest specifying the return type for all these methods to document the API.

interop-cats/shared/src/main/scala/zio/interop/queue/Queue.scala
```diff
@@ -0,0 +1,25 @@
+package zio.interop.queue
+
+import cats.effect.LiftIO
+import zio.{ Runtime, UIO, ZEnv, Queue => ZioQueue }
+
+object Queue {
+  import CQueue._
+
+  type Queue[F[+_], A] = CQueue[F, A, A]
```
adamgfraser: I think we probably want to bring this up into the package object.

interop-cats/shared/src/main/scala/zio/interop/queue/CQueue.scala
calvinlfer: agreed 👍

interop-cats/shared/src/main/scala/zio/interop/queue/CQueue.scala
calvinlfer: done 👍

interop-cats/shared/src/main/scala/zio/interop/queue/CQueue.scala
calvinlfer: factored out 👍

interop-cats/shared/src/main/scala/zio/interop/queue/CQueue.scala
calvinlfer: oops, added 👍

interop-cats/shared/src/main/scala/zio/interop/queue/Queue.scala
calvinlfer: good idea, done 👍

calvinlfer:
Hey, @adamgfraser! 
Thank you for the valuable feedback. I believe I have addressed almost all of your concerns. With regards to tests, I was wondering what you had in mind? So far I have done the following:

```scala
package zio.interop

import cats.effect.{ IO => CIO }
import org.specs2.Specification
import org.specs2.specification.AroundTimeout
import zio.{ DefaultRuntime, ZEnv, ZIO }

class CatsQueueSpec extends Specification with AroundTimeout with DefaultRuntime {
  override def is =
    s2"""
      Queue
      - must be usable from Cats Effect IO $createAndShutdownCIO
      """

  def createAndShutdownCIO =
    unsafeRun {
      ZIO.runtime[ZEnv].flatMap { implicit runtime =>
        ZIO((for {
          q <- Queue.bounded[CIO, Any, Int](10)
          _ <- q.offer(1)
          r <- q.takeAll
          _ <- q.shutdown
        } yield r).unsafeRunSync())
      }
    } must be_===(List(1))
}
```

I was planning to add this to the `zio.interop` package in the `interop-cats/jvm` folder. Could you please give me some guidance on this one?

adamgfraser:
@calvinlfer I would just suggest a few basic tests to demonstrate that the interop package allows working with the queue using an effect type such as Cats IO. Just a few basic tests to demonstrate that: (1) all of the implicits resolve properly so we can actually execute a program and (2) we haven't done something silly like created different queues for each call instead of working with the same queue. Things like offering a couple of values to the queue and getting them, suspending on getting a value from the queue, and maybe using a couple of the effectual `mapM` or `contraMapM` functions. I think I would put it in the shared folder in a new file like `catzQueueSpec.scala` as I think all of this functionality should work across platforms.

calvinlfer:
Thanks @adamgfraser 
Regarding the example test I posted here https://github.com/zio/interop-cats/pull/146#issuecomment-587098479

Is that sufficient machinery to show it working with Cats Effect IO because I just rely on zios DefaultRuntime combined with IOs unsafeRunSync and ZIOs unsafeRun to execute the effect. Is there a better way to do this? 

It’s still unclear to me how to bring in ZIOs runtime into a Cats Effect runtime without wrapping the entire thing in a ZIO effect or using unsafeRun. If I was using Cats Effect as an end user and I wanted to use zios Queue then would I have to extend both IOApp and ZIO.App to get the ZIO.Runtime? Sorry for the trouble

calvinlfer:
oops sorry @adamgfraser I think it was as simple as this: 

```scala
package zio.interop

import cats.effect.{ IO => CIO }
import zio.DefaultRuntime
import zio.test.interop.catz.test._
import zio.test.Assertion._
import zio.test._

object catzQueueSpec
    extends DefaultRunnableSpec(
      suite("catzQueueSpec")(
        testF("Can use Queue from Cats Effect IO provided a zio.Runtime is in scope") {
          implicit val r: zio.Runtime[zio.ZEnv] = new DefaultRuntime {}
          for {
            q <- Queue.bounded[CIO, Any, Int](1)
            _ <- q.offer(1)
            r <- q.takeAll
          } yield assert(r, equalTo(List(1)))
        }
      )
    )
```

calvinlfer:
Hey, @adamgfraser I managed to implement some tests to cover all the cases you mentioned. Hope it looks okay :)

interop-cats/shared/src/test/scala/zio/interop/catzQueueSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+
+import scala.concurrent.ExecutionContext.global
+
+object catzQueueSpec
+    extends DefaultRunnableSpec({
+      def boundedQueueTest[F[+_]](implicit F: Concurrent[F], R: Runtime[ZEnv]): F[TestResult] =
+        for {
+          q  <- Queue.bounded[F, Any, Int](1)
+          _  <- q.offer(1)
+          f  <- q.offer(2).start
```
adamgfraser: This is going to be flaky because the forked fiber could offer the second value to the queue before `takeAll`

calvinlfer: This queue is bounded at size 1 and it’s full after `offer(1)` so it will suspend `offer(2)` and it won’t complete unless a `takeAll` is done. I actually purposely did this to test suspension although I’m not sure if this is the he greatest way to do it

calvinlfer: This queue is bounded at size 1 and it’s full after `offer(1)` so it will suspend `offer(2)` and it won’t complete unless a `takeAll` is done. I actually purposely did this to test suspension although I’m not sure if this is the he greatest way to do it

adamgfraser: Ah, missed that, sorry!

interop-cats/shared/src/main/scala/zio/interop/Queue.scala
```diff
@@ -0,0 +1,50 @@
...
+
+import cats.effect.LiftIO
+import zio.{ Queue => ZioQueue, Runtime, UIO, ZEnv }
+
+object Queue {
+
+  /**
+   * @see [[ZioQueue.bounded]]
+   */
+  final def bounded[F[+_], R, A](capacity: Int)(implicit R: Runtime[ZEnv], F: LiftIO[F]): F[Queue[F, A]] =
```
ghostdogpr: Isn't the `R` type parameter useless here?

interop-cats/shared/src/main/scala/zio/interop/Queue.scala
```diff
@@ -0,0 +1,50 @@
...
+    create(ZioQueue.bounded[A](capacity))
+
+  /**
+   * @see [[ZioQueue.dropping]]
```
ghostdogpr: Using `[[...]]` makes Scaladoc finds the actual type I think? Isn't it going to fail generating the doc? (I'm not sure)

interop-cats/shared/src/main/scala/zio/interop/Queue.scala
adamgfraser: Great point!

interop-cats/shared/src/main/scala/zio/interop/Queue.scala
adamgfraser: Ah I think you're right. Looks like we're not doing the documentation as part of CI for this project. Probably best to do it the same way we did for Schedule I would think.

interop-cats/shared/src/main/scala/zio/interop/Queue.scala
ghostdogpr: Saw it from the tests actually, `Queue.dropping[F, Any, Int](2)` can simply be `Queue.dropping[F, Int](2)`

interop-cats/shared/src/main/scala/zio/interop/Queue.scala
calvinlfer: I actually ran sbt doc against this and it compiled but I can revert to the way Schedule does it

calvinlfer: I actually ran sbt doc against this and it compiled but I can revert to the way Schedule does it

calvinlfer: I actually ran sbt doc against this and it compiled but I can revert to the way Schedule does it

adamgfraser: Maybe look at the generated Scaladoc for this and Schedule and see if the links work?

interop-cats/shared/src/main/scala/zio/interop/Queue.scala
calvinlfer: Yeah you are right 👌

interop-cats/shared/src/main/scala/zio/interop/Queue.scala
calvinlfer: Will do!

interop-cats/shared/src/main/scala/zio/interop/Queue.scala
calvinlfer: Fixed :)

interop-cats/shared/src/main/scala/zio/interop/Queue.scala
calvinlfer: Seems like it was complaining about these and it wasn't for the others so I shifted to the way Schedule does it

calvinlfer:
hey @ghostdogpr and @adamgfraser it should be good to go now 😸

adamgfraser:
Looks great!

PR #156, opened 2020-03-08
Make both sides of race interruptible, generate interruptible races (and fix failing laws that way)

neko-kai:
The tests were failing because we were generating uninterruptible races under `uncancelable` and bracket, which didn't make sense for `eqIO` - 

```scala
-    Gen.const(io.race(IO.never))
+    Gen.const(io.interruptible.race(ZIO.never.interruptible))
```

the cats test `uncancelableMirrorsSource` identified the issue correctly `uncancelable(race) != race` because `race` after RC18 inherits interruption status and becomes completely uncancelable too.

inspired by https://github.com/zio/zio/issues/3065

> **Safer race**
> Race no longer forces either side to be interruptible. Users may have to perform left.interruptible.race(right.interruptible) to acquire the old behavior (which punched holes in uninterruptible regions); or possibly, left.disconnect.race(right.disconnect) (if they want the effects to be keep running when interrupted, but in the background).

### DIFF
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,5 +1,5 @@
 -J-Xss5M
 -J-XX:ReservedCodeCacheSize=256m
--J-Xmx4G
+-J-Xmx5G
 -J-XX:MaxMetaspaceSize=3G
 -J-XX:MaxInlineLevel=18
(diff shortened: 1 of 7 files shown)


neko-kai:
Also SemigroupK law was failing because our `Cause` has become too detailed (i think), so I changed the Eq instance to preserve a bit less information about Throwables and FiberIds

neko-kai:
> because race after RC18 inherits interruption status and becomes completely uncancelable too.

Yes, the `TestContext` is correct, when we generate `race(a, never)`, because `race` always waits for both sides to interrupt, `race(a, never).unacancelable` will never finish, because it will wait forever for `never.uncancelable` to complete which it won't.

PR #164, opened 2020-04-05
Added ArrowChoice instance for ZManaged

eugengarkusha:
discussed with @adamgfraser in zio-contributors


interop-cats/shared/src/main/scala/zio/interop/catszmanaged.scala
```diff
@@ -211,3 +214,33 @@ private class CatsZManagedSync[R] extends CatsZManagedMonadError[R, Throwable] w
...
+  final override def second[A, B, C](fa: ZManaged[A, Any, B]): ZManaged[(C, A), Any, (C, B)] =
+    ZManaged.accessManaged { case (c, a) => fa.provide(a).map((c, _)) }
+  final override def split[A, B, C, D](f: ZManaged[A, Any, B], g: ZManaged[C, Any, D]): ZManaged[(A, C), Any, (B, D)] =
+    ZManaged.accessManaged { case (a, c) => f.provide(a).zip(g.provide(c)) }
+  final override def merge[A, B, C](f: ZManaged[A, Any, B], g: ZManaged[A, Any, C]): ZManaged[A, Any, (B, C)] = f.zip(g)
+
+  final override def lmap[A, B, C](fab: ZManaged[A, Any, B])(f: C => A): ZManaged[C, Any, B] = fab.provideSome(f)
+  final override def rmap[A, B, C](fab: ZManaged[A, Any, B])(f: B => C): ZManaged[A, Any, C] = fab.map(f)
+
+  final override def choice[A, B, C](f: ZManaged[A, Any, C], g: ZManaged[B, Any, C]): ZManaged[Either[A, B], Any, C] =
```
adamgfraser, line 241: I think you should be able to implement most of these by just calling the corresponding methods on `ZManaged`. For example `choice` is just `f ||| g`.

eugengarkusha: I see, thanks, doing!

eugengarkusha: done

interop-cats/shared/src/main/scala/zio/interop/catszmanaged.scala
```diff
@@ -211,3 +214,28 @@ private class CatsZManagedSync[R] extends CatsZManagedMonadError[R, Throwable] w
       }
     }
 }
+
+private object CatsZManagedArrowChoice extends ArrowChoice[ZManaged[*, Any, *]] {
+  final override def lift[A, B](f: A => B): ZManaged[A, Any, B] = ZManaged.fromFunction(f)
+  final override def compose[A, B, C](f: ZManaged[B, Any, C], g: ZManaged[A, Any, B]): ZManaged[A, Any, C] =
+    g >>= f.provide
```
adamgfraser, line 221: This can just be `f compose g`.

interop-cats/shared/src/main/scala/zio/interop/catszmanaged.scala
```diff
@@ -211,3 +214,28 @@ private class CatsZManagedSync[R] extends CatsZManagedMonadError[R, Throwable] w
...
+
+  final override def id[A]: ZManaged[A, Any, A] = ZManaged.environment
```
adamgfraser, line 223: Possibly slightly more idiomatically `ZManaged.identity`.

interop-cats/shared/src/test/scala/zio/interop/catzSpecBase.scala
```diff
@@ -50,4 +51,24 @@ private[interop] sealed trait catzSpecBaseLowPriority { this: catzSpecBase =>
     Eq.instance((io1, io2) => Arbitrary.arbitrary[R].sample.fold(false)(r => catsSyntaxEq(run(r, io1)) eqv run(r, io2)))
   }
 
+  //---!! TODO: this is the comment for a reviewer: remove or rewrite it after the review!!---
```
adamgfraser, line 54: We've got to figure out a way to organize the instances so we don't have these issues? Maybe we can mirror the organization structure for the `ZIO` instances, which seem to be working for instances including ones with an environment type.

eugengarkusha: Fixed. Thanks for your advise.
The problem is really strange because previously Eq for ZManaged[Any, E, A] was even in higher priority scope then now

adamgfraser:
@ghostdogpr or @neko-kai Would you mind taking a look at this? The type class instances seem very brittle where adding a new `Eq` instance for `ZManaged` breaks things but I am not a Cats expert so maybe there is something obvious I am missing here. Alternatively we could merge as is and open a new issue to try to clean this up.

neko-kai:
@adamgfraser @eugengarkusha 
I've tried cleaning it up, but I didn't really get to why the implicits diverge. I've opened a PR for your, @eugengarkusha, branch. https://github.com/eugengarkusha/interop-cats/pull/1
Behavior of pre-2.13 scala doesn't really make much sense to me here, I think it works on 2.13+ just as a result of optimizations for implicit search there (Seems like previous Scalas consider all polymorphic implicits first, even if their output type constructors are unrelated to the searched type, etc.)

adamgfraser:
@neko-kai Thank you! 🙏

eugengarkusha:
@neko-kai thanks for your suggestion!
I still have one question: 
Do you want me to merge your branch as is or you confirm the current state of this PR?

Asking because the changes in your branch are more verbose and since this fix is only related to older version of compiler, maybe it makes sense to fix it in less invasive way?

Thanks!

neko-kai:
@eugengarkusha Yeah i think let's merge as-is once ci finishes

PR #191, opened 2020-06-09
Add NonEmptyChunk reducible instance

renancaju:
@adamgfraser, could you give me some feedback on wether this is acceptable?

I was told `Chunk#foldRight` isn't lazy enough, so implementing `foldRight` and `reduceRightTo` in terms of it would not be optimal (I don't know if this is critical though).

Regarding `reduceRightTo`, I couldn't find a way to pattern match on the `NonEmptyChunk` so as to not use `last` (which is unsafe) so I followed the pragmatical approach with the `if / else` expression.

I'd be happy work some more on this if my solution is not adequate.

### DIFF
--- a/interop-cats/shared/src/main/scala/zio/interop/cats.scala
+++ b/interop-cats/shared/src/main/scala/zio/interop/cats.scala
@@ -121,6 +121,8 @@ abstract class CatsInstances extends CatsInstances1 {
   implicit final def contravariantInstance[E, A]: Contravariant[ZIO[*, E, A]] =
     contravariantInstance0.asInstanceOf[Contravariant[ZIO[*, E, A]]]
 
+  implicit final def nonEmptyChunkReducibleInstance: Reducible[NonEmptyChunk] = new NonEmptyChunkReducible
+
   private[this] val bifunctorInstance0: Bifunctor[ZIO[Any, *, *]]           = new CatsBifunctor
   private[this] val zioArrowInstance0: ArrowChoice[ZIO[*, Any, *]]          = new CatsArrow
   private[this] val contravariantInstance0: Contravariant[ZIO[*, Any, Any]] = new CatsContravariant
@@ -384,4 +386,19 @@ private class CatsArrow[E] extends ArrowChoice[ZIO[*, E, *]] {
 final private class CatsContravariant[E, T] extends Contravariant[ZIO[*, E, T]] {
   override def contramap[A, B](fa: ZIO[A, E, T])(f: B => A): ZIO[B, E, T] =
     ZIO.accessM[B](b => fa.provide(f(b)))
+}
+
+final private class NonEmptyChunkReducible extends Reducible[NonEmptyChunk] {
+
+  override def foldLeft[A, B](fa: NonEmptyChunk[A], b: B)(f: (B, A) => B): B = fa.foldLeft(b)(f)
+  override def foldRight[A, B](fa: NonEmptyChunk[A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
+    fa.foldRight(lb)(f)
+
+  override def reduceLeftTo[A, B](fa: NonEmptyChunk[A])(f: A => B)(g: (B, A) => B): B =
+    fa.tail.foldLeft(f(fa.head))(g)
+
+  override def reduceRightTo[A, B](fa: NonEmptyChunk[A])(f: A => B)(g: (A, Eval[B]) => Eval[B]): Eval[B] =
+    if (fa.tail.isEmpty) Eval.later(f(fa.head))
+    else fa.foldRight(Eval.later(f(fa.last)))(g)
+
 }
--- a/interop-cats/shared/src/main/scala/zio/interop/cats.scala
+++ b/interop-cats/shared/src/main/scala/zio/interop/cats.scala
@@ -122,7 +122,7 @@ abstract class CatsInstances extends CatsInstances1 {
     contravariantInstance0.asInstanceOf[Contravariant[ZIO[*, E, A]]]
 
   implicit final val nonEmptyChunkReducibleInstance: Reducible[NonEmptyChunk] = new NonEmptyChunkReducible
-  implicit final def nonEmptyChunkEqInstance[A]: Eq[NonEmptyChunk[A]]          = new NonEmptyChunkEq[A]
+  implicit final def nonEmptyChunkEqInstance[A]: Eq[NonEmptyChunk[A]]         = new NonEmptyChunkEq[A]
 
   private[this] val bifunctorInstance0: Bifunctor[ZIO[Any, *, *]]           = new CatsBifunctor
   private[this] val zioArrowInstance0: ArrowChoice[ZIO[*, Any, *]]          = new CatsArrow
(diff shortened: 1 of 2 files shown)


interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -121,6 +121,8 @@ abstract class CatsInstances extends CatsInstances1 {
   implicit final def contravariantInstance[E, A]: Contravariant[ZIO[*, E, A]] =
     contravariantInstance0.asInstanceOf[Contravariant[ZIO[*, E, A]]]
 
+  implicit final def nonEmptyChunkReducibleInstance: Reducible[NonEmptyChunk] = new NonEmptyChunkReducible
```
adamgfraser, line 124:
```suggestion
  implicit final val nonEmptyChunkReducibleInstance: Reducible[NonEmptyChunk] = new NonEmptyChunkReducible
```

interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -385,3 +387,18 @@ final private class CatsContravariant[E, T] extends Contravariant[ZIO[*, E, T]]
...
+final private class NonEmptyChunkReducible extends Reducible[NonEmptyChunk] {
+
+  override def foldLeft[A, B](fa: NonEmptyChunk[A], b: B)(f: (B, A) => B): B = fa.foldLeft(b)(f)
+  override def foldRight[A, B](fa: NonEmptyChunk[A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
+    fa.foldRight(lb)(f)
+
+  override def reduceLeftTo[A, B](fa: NonEmptyChunk[A])(f: A => B)(g: (B, A) => B): B =
+    fa.tail.foldLeft(f(fa.head))(g)
+
+  override def reduceRightTo[A, B](fa: NonEmptyChunk[A])(f: A => B)(g: (A, Eval[B]) => Eval[B]): Eval[B] =
```
adamgfraser, line 400: Why is `last` unsafe? The `NonEmptyChunk` is guaranteed to be non-empty so there should always be a last element.

interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -385,3 +387,18 @@ final private class CatsContravariant[E, T] extends Contravariant[ZIO[*, E, T]]
   override def contramap[A, B](fa: ZIO[A, E, T])(f: B => A): ZIO[B, E, T] =
     ZIO.accessM[B](b => fa.provide(f(b)))
 }
+
+final private class NonEmptyChunkReducible extends Reducible[NonEmptyChunk] {
+
+  override def foldLeft[A, B](fa: NonEmptyChunk[A], b: B)(f: (B, A) => B): B = fa.foldLeft(b)(f)
+  override def foldRight[A, B](fa: NonEmptyChunk[A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
```
adamgfraser, line 394: `Chunk#foldRight` could potentially be lazier but this should not matter here as we are folding using `Eval` which supports lazy evaluation.

interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -385,3 +388,39 @@ final private class CatsContravariant[E, T] extends Contravariant[ZIO[*, E, T]]
...
+
+}
+
+final private class NonEmptyChunkEq[A] extends Eq[NonEmptyChunk[A]] {
+
+  override def eqv(x: NonEmptyChunk[A], y: NonEmptyChunk[A]): Boolean =
+    if (x.length != y.length) false
+    else x.zip(y).forall { case (a, b) => a == b }
+
+}
```
renancaju, line 426: The law specs needed an `Eq` instance as well

interop-cats/shared/src/test/scala/zio/interop/catzSpec.scala
```diff
@@ -41,6 +42,28 @@ class catzSpec extends catzSpecZIOBase {
...
+          head <- gen
+          tail <- Gen.listOf(gen)
+        } yield NonEmptyList(head, tail)
+
+      def nonEmptyChunk[A](gen: => Gen[A]): Gen[NonEmptyChunk[A]] =
+        nonEmptyListGen(gen).map(nel => NonEmptyChunk.fromCons(::(nel.head, nel.tail)))
+
+      implicit val arbitratyNonEmptyChunk: Arbitrary[NonEmptyChunk[Int]] = Arbitrary(nonEmptyChunk(intGen))
+      implicit val arbitratyNonEmptyChunkOpt: Arbitrary[NonEmptyChunk[Option[Int]]] =
+        Arbitrary(nonEmptyChunk(Gen.option(intGen)))
```
renancaju, line 61: Both `Arbitrary`s were needed for the `ReducibleTests`

interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -385,3 +388,39 @@ final private class CatsContravariant[E, T] extends Contravariant[ZIO[*, E, T]]
...
+      else loop(idx - 1, Eval.defer(g(fa(idx), acc)))
+
+    Eval.defer(loop(idx, Eval.now(acc)))
+  }
+
+}
+
+final private class NonEmptyChunkEq[A] extends Eq[NonEmptyChunk[A]] {
+
+  override def eqv(x: NonEmptyChunk[A], y: NonEmptyChunk[A]): Boolean =
```
adamgfraser, line 422: I think we want to require an `Eq[A]` here and use that to compare pairwise elements instead of universal equality.

interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -385,3 +388,39 @@ final private class CatsContravariant[E, T] extends Contravariant[ZIO[*, E, T]]
   override def contramap[A, B](fa: ZIO[A, E, T])(f: B => A): ZIO[B, E, T] =
     ZIO.accessM[B](b => fa.provide(f(b)))
 }
+
+final private class NonEmptyChunkReducible extends Reducible[NonEmptyChunk] {
+
+  override def foldLeft[A, B](fa: NonEmptyChunk[A], b: B)(f: (B, A) => B): B = fa.foldLeft(b)(f)
+  override def foldRight[A, B](fa: NonEmptyChunk[A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] = {
```
adamgfraser, line 395: This is going to have to traverse the entire chunk to build up the `Eval` data structure even though we may only need to evaluate part of it. Instead you can do:

```scala
  override def foldRight[A, B](fa: NonEmptyChunk[A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] = {
    val length = fa.length
    def loop(n: Int): Eval[B] =
      if (n == length) lb
      else f(fa(n), Eval.defer(loop(n + 1)))

    Eval.defer(loop(0))
  }
```

Can make a similar change with `reduceLeftTo` below.

neko-kai: @renancaju I think this comment is the only thing preventing a merge currently? ☝️

renancaju: To tell the truth, I didn't do it because I didn't fully understand Adam's suggestion. Instead of just copying and pasting what he said, I decided this feature should be in more experienced hands. Please, feel free to close this PR and implement a better one from scratch. Sorry I didn't say that before!

PR #195, opened 2020-06-20
Update ZIO to 1.0.0-RC21

ghostdogpr:


interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -201,8 +201,8 @@ private class CatsConcurrentEffect[R](rts: Runtime[R])
             _ => fa
           )
           .interruptible
-          .forkWithErrorHandler(_ => ZIO.unit)
-          .tap(ZIO.disown)
+          .onError(_.failureOrCause.fold(_ => ZIO.unit, ZIO.halt(_)))
```
neko-kai, line 204: this won't suppress the error because onError does not handle errors

neko-kai: @adamgfraser is there a more direct replacement for `forkWithErrorHandler` in RC21?

ghostdogpr: `forkWithErrorHandler` is still there, it’s `disown` that’s gone. Ire placed it with `forkDaemon` but couldn’t use `forkWithErrorHandler as it does a normal fork. The `onError` thing is just copied from the implementation of `forkWithErrorHandler`: https://github.com/zio/zio/blob/556af9d03011ec18666883566cd89c850636f27b/core/shared/src/main/scala/zio/ZIO.scala#L789

ghostdogpr: Actually, the logic to not trigger the default reporter (`.run.fork.map(_.mapM(IO.done))`) was removed in this PR: https://github.com/zio/zio/pull/2715/files#diff-9cfb50d60ba29ffb4a6a15c933dbf2c3L629 @adamgfraser any idea why?

I think that ever since, we were not bypassing the default error reporter, but since it only logs `Die` errors now, no one noticed.

ghostdogpr: So here I can either add `.run.forkDaemon.map(_.mapM(IO.done))`, or just remove that line. Wdyt?

ghostdogpr: @neko-kai applied the trick to prevent the interruption from being reported. I think we should fix `forkWithErrorHandler` in ZIO core since that behavior was lost.

ghostdogpr: After discussing with @adamgfraser, I just removed `.run`/`.mapM(IO.done)`. The behavior will be same as RC20, which is that only `Die` errors are logged in the default reporter.

Issue #69, opened 2019-09-22
Create an interop-fs2 project for ZIO Streams


iravid:
Opening this ticket here just for tracking; project should probably be created in a separate repository.

Provide conversions from `ZStream` to `fs2.Stream` and from `fs2.Stream` to `ZStream`.

cc @fsvehla

ghostdogpr:
Transferred the issue to interop-cats

iravid:
Implementation for `fs2.Stream => ZStream`, if anyone wants to pick this up for a PR:
```scala
  implicit class Fs2Stream[R, A](stream: fs2.Stream[RIO[R, ?], A]) {
    def toZStream: ZStream[R, Throwable, A] =
      ZStream
        .managed {
          for {
            queue <- Queue.bounded[Take[Throwable, A]](1).toManaged(_.shutdown)
            _ <- ZIO
                  .runtime[R]
                  .toManaged_
                  .flatMap { implicit runtime =>
                    stream
                      .evalTap(a => queue.offer(Take.Value(a)))
                      .onFinalizeCase {
                        case ExitCase.Canceled  => queue.offer(Take.Fail(Cause.interrupt)).unit
                        case ExitCase.Completed => queue.offer(Take.End).unit
                        case ExitCase.Error(e)  => queue.offer(Take.Fail(Cause.fail(e))).unit
                      }
                      .compile
                      .resource
                      .drain
                      .toManaged
                  }
                  .fork
          } yield ZStream.fromQueue(queue).unTake
        }
        .flatMap(identity)
  }
```

adamw:
That would be quite usefull for ztapir :) I could then add a way to translate raw streams and we'd have streaming support using http4s. Any chance somebody has an updated version of the above mapping? :)

PR #203, opened 2020-07-07
Issue #69 - add fs2.Stream conversion

mindartur:
Modified version of proposed solution by @iravid in #69

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -53,7 +53,7 @@ lazy val interopCats = crossProject(JSPlatform, JVMPlatform)
       "dev.zio"       %%% "zio-test"             % zioVersion % Optional,
       "org.typelevel" %%% "cats-effect"          % "2.1.3" % Optional,
       "org.typelevel" %%% "cats-mtl-core"        % "0.7.1" % Optional,
-      "co.fs2"        %%% "fs2-core"             % "2.3.0" % Test,
+      "co.fs2"        %%% "fs2-core"             % "2.3.0" % Optional,
       "dev.zio"       %%% "zio-test-sbt"         % zioVersion % Test,
       "org.specs2"    %%% "specs2-core"          % "4.9.4" % Test,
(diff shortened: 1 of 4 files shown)


ghostdogpr:
Looks OK to me. @iravid you wanna double check?

iravid:
This version looks good to merge to me. Great coverage @mindartur and love the chunk variant.

As a follow-up, I would suggest *always* using chunks, and maintaining the original chunking from the `fs2.Stream`. That is, use `evalTapChunk` (I think that exists - if not, then `evalMapChunk`), convert the `fs2.Chunk` to `zio.Chunk` and enqueue that.

interop-cats/shared/src/main/scala/zio/stream/interop/fsz2.scala
```diff
@@ -0,0 +1,5 @@
+package zio.stream.interop
+
+object fsz2 extends FS2Platform
```
ghostdogpr, line 3: Btw shouldn't it be fs2z? 😄

mindartur: yeah, right, I forgot it is scala**Z** :D

mindartur:
@iravid @ghostdogpr Also, do you think that `toZStream(): ZStream[R, Throwable, A]` should always return `ZStream[Any, Throwable, A]`? 

I mean, `fs2.Stream` can't have any environment, right?

ghostdogpr:
If you force the fs2 Stream to be on `Task` yeah the R can be just Any. But can’t you change it to `RIO[R, *]` so that it works for any R?

mindartur:
> If you force the fs2 Stream to be on `Task` yeah the R can be just Any. But can’t you change it to `RIO[R, *]` so that it works for any R?

Changed `Task` to `RIO`

PR #207, opened 2020-07-14
remove specs2

### LINKED ISSUE #206: Port tests using spec2 to zio test
Example tests using spec2: https://github.com/zio/interop-cats/blob/master/interop-cats/jvm/src/test/scala/zio/interop/CatsInteropSpec.scala#L4

lemastero:
fix #206


interop-cats/jvm/src/test/scala/zio/interop/CatsZManagedSyntaxSpec.scala
```diff
@@ -3,334 +3,302 @@ package zio.interop
...
+            managed.use(_ => ZIO.fail(new RuntimeException()).unit)
+          }
+
+          unsafeRun(testCase.orElse(ZIO.unit))
+          assert(effects.toList)(equalTo(List(1, 2)))
+        },
+        test("calls finalizers correctly when use has died") {
+          val effects = new mutable.ListBuffer[Int]
+          def res(x: Int): Resource[CIO, Unit] =
+            Resource.makeCase(CIO.delay { effects += x; () }) {
```
olownia, line 60: `effects += x; ()` is really ugly :)

jczuchnowski: Probably could be replaced by `.as(())`

neko-kai: `.void`. But I actually prefer `; ()` in general since it's more efficient.

lemastero: Thank you for review 🙇 
Changed to `.void` / `.unit`
(Since this is in test code, it is probably more beneficial to have code easy to read than tuned in terms of performance.)

interop-cats/jvm/src/test/scala/zio/interop/Fs2ZioSpec.scala
```diff
@@ -4,113 +4,98 @@ package interop
...
+              terminate.await *> IO.die(new Exception())
+            }
+            .compile
+            .drain
+            .fork
+
+      _ <- started.await
+      _ <- terminate.succeed(())
+      _ <- released.await
+    } yield assert(())(anything)
```
olownia, line 82: is this assertion make any sense? shouldn't we test results?

lemastero: @molowny

Replaced `assert(())(anything)` with `assertCompletes` as there is already a nice combinator to achieve such check.

The idea of this test is to check if things "work when fiber is interrupted". If there is an error somewhere, it will be encoded in ZIO and found by zio-test. (At least that is my understanding.)

neko-kai:
@lemastero Is this ready for merge? Thanks!

lemastero:
Thanks for reviewers for their time and remarks 🙇 
Yes! Let's 🚢  it!

PR #212, opened 2020-08-04
Fix nested race issue

ghostdogpr:
Fixes https://github.com/zio/zio/issues/4024

In the last release we added `overrideForkScope(ZScope.global)` to make sure anything forked by a side of the race would be global. The memory leak happens when there is a nested race, because the forks internal to race should not be global. 
By calling `resetForkScope` we ensure those forks are "normal" and only the ones inside left and right are global.

It shows no more leak with the reproducer, and the http4s scope issue doesn't happen either.

I also removed `race` because cats-effect already has an implementation for it using `racePair`.

cc @Fristi @sebver

### DIFF
--- a/interop-cats/shared/src/main/scala/zio/interop/cats.scala
+++ b/interop-cats/shared/src/main/scala/zio/interop/cats.scala
@@ -227,13 +227,6 @@ private class CatsConcurrent[R] extends CatsMonadError[R, Throwable] with Concur
       Left(token.orDie)
     }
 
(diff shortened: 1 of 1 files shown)


sebver:
I can confirm the memory leak is solved and the http4s example is also still working.

ghostdogpr:
Great to hear! I'll merge then. We can release once the upgrade to ZIO 1.0.0 is ready.

PR #217, opened 2020-08-12
Use docker image with sbt on CI

joroKr21:


joroKr21:
We use docker on CircleCI at work every day with no issues but it's up to you ofc.

joroKr21:
> We use docker on CircleCI at work every day with no issues but it's up to you ofc.

Funny enough, we had issues with machine that went away when we switched to Docker 😂

PR #233, opened 2020-09-23
Update zio 1.0.1 fix

lemastero:
PR #222 cannot be automatically merged. This should solve the problems.


interop-cats/shared/src/main/scala/zio/interop/Schedule.scala
```diff
@@ -159,7 +159,7 @@ final class Schedule[F[+_], -In, +Out] private (private[Schedule] val underlying
   def combineWith[In1 <: In, Out2](
```
adamgfraser, line 159: In ZIO we deprecated `combineWith` and added the `unionWith` and `intersectWith` operators. I would suggest we do the same here to maintain a consistent API.

adamgfraser (changes requested):
Looks great! Just one suggestion on `Schedule`. Thanks for doing this!

interop-cats/shared/src/main/scala/zio/interop/Schedule.scala
lemastero: Thank you for the review. Deprecated `combineWith` in favour of `intersectWith ` and added `unionWith` as well.

adamgfraser (approved):
Looks great! Thanks again for working on this! 🙏

PR #234, opened 2020-10-03
Fix resource leak in `.toResourceZIO` that could happen just after ZManaged was acquired but before the finalizer was registered in Resource interpreter (https://github.com/typelevel/cats-effect/blob/bf1fa530bd651325c7b060389952d37b4852d5e5/core/shared/src/main/scala/cats/effect/Resource.scala#L133)

neko-kai:
Basically, there's a `flatMap` in-between a `Resource.Suspend` node ending and `Resource.Allocate` node being processed (- https://github.com/typelevel/cats-effect/blob/bf1fa530bd651325c7b060389952d37b4852d5e5/core/shared/src/main/scala/cats/effect/Resource.scala#L133)
If Resource produced by `toResourceZIO` was interrupted at that point, the bracketCase's inside `managed.zio` would already finish, but the finalizer for the `resourceMap` wouldn't be registered yet, because it's in the `Resource.Allocate` node which is the result of `Suspend`. This moves the `Resource.Allocate` node _before_ suspend, so that the finalizer for the `releaseMap` is registered before the `managed.zio` is evaluated and the finalizers are registered as promptly as possible.

### DIFF
--- a/interop-cats/shared/src/main/scala/zio/interop/catszmanaged.scala
+++ b/interop-cats/shared/src/main/scala/zio/interop/catszmanaged.scala
@@ -17,12 +17,10 @@
 package zio
 package interop
 
-import cats.~>
-import cats.arrow.FunctionK
+import cats.arrow.{ ArrowChoice, FunctionK }
 import cats.effect.Resource.{ Allocate, Bind, Suspend }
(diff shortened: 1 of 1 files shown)


ghostdogpr (approved):
Good catch 👍

PR #244, opened 2020-10-27
Support Converting a ZIO Clock Into A Cats Effect Timer

### LINKED ISSUE #243: Compatibility with TestClock
While writing tests for ZIO-wrapped fs2.Stream code, I discovered that using `TestClock.adjust` or similar has no effect on the `Timer[Task]` instance made available through `zio.interop.catz.implicits`.

A reproducer (though Scastie doesn't let me choose any newer version of the interop package than 1.0.3, so it won't run): https://scastie.scala-lang.org/q1qeLLL9Rty3E0NPT0izGQ

@adamgfraser had this to say:
Yes I think that is an issue. The problem seems to be that in `﻿interop-cats﻿` the `﻿Timer﻿` instance uses the live `﻿Clock﻿` so like in your example the stream you are creating using the live clock. This makes sense in general because the Cats Effect types can't require an environment so we need to provide something and the live one makes the most sense, though it doesn't work for your use case. I think what we could probably do there is create a helper method to create a Cats Effect `﻿Timer﻿` and `﻿Clock﻿` from a ZIO `﻿Clock﻿`. Then you could access the `﻿Clock﻿` in the environment, which would be the `﻿TestClock﻿`, and use it to define a new implicit `﻿Timer﻿` instance that would use that. Do you want to open an issue for that in `﻿interop-cats﻿`?

adamgfraser:
Resolves #243.

Our default `Timer` instead just uses the live `Clock`, which makes sense because we don't have the ability to further defer providing an implementation and this is most likely to be the correct one. However, it raises an issue when a user wants to use a `Timer` backed by a different `Clock` implementation, such as the `TestClock`.

This PR addresses the issue by defining a new `toTimer` operator on `Clock.Service` available through implicit syntax. With this a user who wants to define a `Timer` backed by a different ZIO `Clock` can just define a new implicit `Timer` using this operator to override the instance backed by the live `Clock`.


interop-cats/shared/src/test/scala/zio/interop/catzClockSpec.scala
```diff
@@ -0,0 +1,34 @@
...
+import zio.test.environment._
+import zio.test._
+
+import scala.concurrent.duration.FiniteDuration
+
+object catzClockSpec extends DefaultRunnableSpec {
+
+  def spec =
+    suite("catzClockSpec") {
+      testM("Timer can be constructed from ZIO Clock") {
```
gannicottb, line 18: This test relies on someone noticing that it ran for 20 seconds to determine whether it failed - would it be appropriate to use a `timeout` aspect here, or is there a more elegant way to make sure that the TestClock.adjust was able to skip over the fs2.Stream.delayBy?

neko-kai (approved):
LGTM, but the test should probably use a `MockClock`

interop-cats/shared/src/test/scala/zio/interop/catzClockSpec.scala
adamgfraser: @gannicottb Good point! Added a timeout so the test will fail if we aren't actually using the `TestClock`.

neko-kai (approved):

gannicottb:
Hey, just popping in to say I'm looking forward to this feature!

PR #263, opened 2020-12-01
Port for CE3

joroKr21:
For now just the instances.
Tests will be a pain 🙈


joroKr21:
> Tests will be a pain 🙈

Waiting for typelevel/cats-effect#1556

kubukoz:
Just dropping by to mention typelevel/cats=effect#1556 is merged :)

interop-cats/shared/src/test/scala/zio/interop/GenIOInteropCats.scala
```diff
@@ -48,6 +48,9 @@ trait GenIOInteropCats {
   def genIO[E: Arbitrary, A: Arbitrary]: Gen[IO[E, A]] =
     Gen.oneOf(genSuccess[E, A], genFailure[E, A])
 
+  def genUIO[A: Arbitrary]: Gen[UIO[A]] =
```
adamgfraser, line 51: `genOfRace` below needs to be changed to:

```scala
  private def genOfRace[E, A](io: IO[E, A]): Gen[IO[E, A]] =
    Gen.const(io.raceFirst(ZIO.never))
```

The current implementation in terms of `race` means we wait for the second fiber if the first fiber fails, which means if the first fiber fails we suspend indefinitely.

interop-cats/shared/src/test/scala/zio/interop/catzSpecBase.scala
```diff
@@ -23,59 +26,123 @@ private[zio] trait catzSpecBase
...
+    case (Exit.Success(x), Exit.Success(y)) => x eqv y
+    case (Exit.Failure(x), Exit.Failure(y)) => x eqv y
+    case _                                  => false
+  }
 
-  implicit def zioEqUIO[A: Eq](implicit rts: Runtime[Any], tc: TestContext): Eq[UIO[A]] =
-    Eq.by(uio => taskEffectInstance.toIO(uio.sandbox.either))
+  implicit def eqForUIO[A: Eq](implicit ticker: Ticker): Eq[UIO[A]] = Eq.by { uio =>
+    try {
+      var exit: Exit[Nothing, A] = null
```
adamgfraser, line 90: You need to return an `Exit[Nothing, Option[A]]` here and set it to `None` initially since some tests check for no value being returned after the `TestContext` completes execution and `null` is not equal to `null`. Something like:

```scala
  def unsafeRun[A](zio: UIO[A])(implicit ticker: Ticker): Exit[Nothing, Option[A]] =
    try {

      var exit: Exit[Nothing, Option[A]] = Exit.succeed(None)

      runtime(ticker).unsafeRunAsync {
        zio
          .flatMap(ZIO.succeed(_))
          .on(ticker.ctx)
      } {
        case Exit.Success(a)     => exit = Exit.Success(Some(a))
        case Exit.Failure(cause) => exit = Exit.halt(cause)
      }

      ticker.ctx.tickAll(FiniteDuration(1L, SECONDS))

      exit
    } catch {
      case t: Throwable =>
        t.printStackTrace()
        throw t
    }

  implicit def eqForUIO[A: Eq](implicit ticker: Ticker): Eq[UIO[A]] = Eq.instance { (left, right) =>
    val l = unsafeRun(left)
    val r = unsafeRun(right)
    val p = l eqv r
    if (!p) println(s"$l was not equal to $r")
    p
  }
```

interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -35,348 +38,443 @@ abstract class CatsEffectPlatform
...
+    ZIO.yieldNow
+
+  override final def forceR[A, B](fa: F[A])(fb: F[B]): F[B] =
+    fa.either *> fb
+
+  override final def uncancelable[A](body: Poll[F] => F[A]): F[A] =
+    ZIO.uninterruptibleMask(body.compose(toPoll))
+
+  override final val canceled: F[Unit] =
+    ZIO.halt(Cause.interrupt(Fiber.Id.None))
```
adamgfraser, line 204: There is a significant difference between the ZIO and Cats Effect 3 interruption model here that I am not sure how we handle. In Cats Effect 3 self-interruption is ignored in an uninterruptible region, so we would have to check the interrupt status here and return `ZIO.interrupt` only if we are in an interruptible region. However, under the Cats Effect 3 model self interruption in an uninterruptible region is still observed at the end of the region which doesn't really make sense in the ZIO model because we treat an interruption as a value that can be handled with operators such as `run`.

interop-cats/jvm/src/main/scala/zio/interop/ZioAsync.scala
```diff
@@ -0,0 +1,59 @@
...
+import zio.{ RIO, ZIO }
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+private class ZioAsync[R <: Clock with Blocking] extends ZioTemporal[R, Throwable] with Async[RIO[R, *]] {
+
+  override final def evalOn[A](fa: F[A], ec: ExecutionContext): F[A] =
+    ZIO.effectSuspendTotalWith { (platform, _) =>
+      fa.lock(Executor.fromExecutionContext(platform.executor.yieldOpCount)(ec))
+    }
```
adamgfraser, line 17:
```suggestion
  fa.on(ec)
```

interop-cats/jvm/src/main/scala/zio/interop/ZioAsync.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  override final val executionContext: F[ExecutionContext] =
+    ZIO.effectSuspendTotalWith { (platform, _) =>
+      ZIO.succeedNow(platform.executor.asEC)
+    }
```
adamgfraser, line 22:
```suggestion
    ZIO.executor.map(_.asEC)
```

interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -35,348 +38,443 @@ abstract class CatsEffectPlatform
...
+  override def ref[A](a: A): F[effect.Ref[F, A]] =
+    ZRef.make(a).map(new ZioRef(_))
+
+  override def deferred[A]: F[Deferred[F, A]] =
+    Promise.make[E, A].map(new ZioDeferred(_))
 
-  override final def toIO[A](fa: RIO[R, A]): effect.IO[A] =
-    effect.ConcurrentEffect.toIOFromRunCancelable(fa)(this)
+  override final def start[A](fa: F[A]): F[effect.Fiber[F, E, A]] =
+    fa.fork.map(toFiber)
```
adamgfraser, line 189:
```suggestion
    fa.interruptible.fork.map(toFiber)
```
I believe this is another difference between the ZIO and Cats Effect 3 models. In ZIO interruptibility is regional and is inherited by forked fibers, so if a fiber is forked in an uninterruptible region it will itself be uninterruptible. In contrast my understanding is that in Cats Effect 3 fibers are interruptible when they are forked.

You can see that in a test like this:

```scala
  def uncancelableStartIsCancelable =
    F.uncancelable(_ => F.start(F.never[Unit]).flatMap(f => f.cancel >> f.join)) <-> F.pure(
      Outcome.Canceled())
```

The fiber is forked in an uninterruptible region so I would expect an attempt to interrupt it to suspend indefinitely but this test seems to expect that instead the interruption will successfully complete.

adamgfraser:
@joroKr21 Left some comments above. With these changes I am seeing 14 to 15 test failures. The law failures are:

- [ ] GenSpawn[IO[Int, *]].concurrent.race derives from racePair (left)
- [ ] GenSpawn[IO[Int, *]].concurrent.canceled sequences onCancel in order
- [ ] Async[Task].async.suspend throw is raiseError
- [ ] Async[Task].async.repeated suspend is not memoized
- [ ] Async[Task].async.race derives from racePair (left)
- [ ] Async[Task].async.executionContext commutativity
- [ ] Async[Task].async.evalOn local pure
- [ ] Async[Task].async.canceled sequences onCancel in order
- [ ] Async[Task].async.async right is uncancelable sequenced pure

I think the ones with `racePair` are related to the point I made above regarding differences in the interruption model. A couple of the ones with `evalOn` and `executionContext` look like issues with the equality definition of `ExecutionContext` since ZIO is implemented in terms of an `Executor` and creates new `ExecutionContext` interfaces on demand backed by the same underlying thread pool. Not sure what is going on with the `suspend` tests.

interop-cats/shared/src/test/scala/zio/interop/catzSpecBase.scala
```diff
@@ -23,59 +26,123 @@ private[zio] trait catzSpecBase
     with TestInstances
     with catzSpecBaseLowPriority {
 
-  type Env = Clock with Console with System with Random
+  def checkAllAsync(name: String, f: Ticker => Laws#RuleSet): Unit =
+    checkAll(name, f(Ticker()))
+
+  def environment(implicit ticker: Ticker): ZEnv = {
```
adamgfraser, line 32: On the JVM you need to create a test `Blocking` service backed by the `ticker`. Otherwise tests involving blocking effects will fail because the `TestContext` won't see them.

adamgfraser:
Updated based on using a version of `Blocking` backed by the `TestContext`. Down to six law failures four of which are really the same issue around differences in the interruption model. The other two I think will be resolved by using an appropriate definition of equality for `ExecutionContext`.

interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -35,348 +38,443 @@ abstract class CatsEffectPlatform
...
+    fa.either *> fb
+
+  override final def uncancelable[A](body: Poll[F] => F[A]): F[A] =
+    ZIO.uninterruptibleMask(body.compose(toPoll))
+
+  override final val canceled: F[Unit] =
+    ZIO.halt(Cause.interrupt(Fiber.Id.None))
+
+  override final def onCancel[A](fa: F[A], fin: F[Unit]): F[A] =
+    fa.onInterrupt(fin.orDieWith(fiberFailure))
```
adamgfraser, line 207:
```suggestion
    fa.onInterrupt(fin.ignore)
```
Per laws failure must be suppressed here. 😱

interop-cats/jvm/src/main/scala/zio/interop/ZioAsync.scala
```diff
@@ -0,0 +1,59 @@
...
+  override final def defer[A](thunk: => F[A]): F[A] =
+    ZIO.effectSuspend(thunk)
+
+  override final def blocking[A](thunk: => A): F[A] =
+    effectBlocking(thunk)
+
+  override final def interruptible[A](many: Boolean)(thunk: => A): F[A] =
+    effectBlockingInterrupt(thunk)
+
+  override final def async[A](k: (Either[Throwable, A] => Unit) => F[Option[F[Unit]]]): F[A] =
```
adamgfraser, line 48: I think we need a slightly more completed state machine here to honor the guarantee expected by Cats Effect that if the `F` returned by the register function fails the return value of the callback will not be observed. Something like:

```scala
    Promise.make[Nothing, Unit].flatMap { promise =>
      ZIO.effectAsyncM { register =>
        k(either => register(promise.await *> ZIO.fromEither(either))) *> promise.succeed(())
      }
    }
```

adamgfraser:
@joroKr21 Submitted a pull request to your branch. Everything passing now except for two FS2 tests.

interop-cats/shared/src/test/scala/zio/interop/catzSpecBase.scala
joroKr21: Ah that makes sense, thanks 👍

interop-cats/shared/src/main/scala/zio/interop/cats.scala
joroKr21: I must admit I don't quite understand this.

> which doesn't really make sense in the ZIO model because we treat an interruption as a value that can be handled with operators such as run.

Why can't the same be done in CE3? It also has `Outcome.Canceled`

interop-cats/shared/src/main/scala/zio/interop/cats.scala
joroKr21: Ah I was missing that, thanks 👍 - I think we also need `uninterruptible` in `onCancel` - WDYT?

interop-cats/shared/src/main/scala/zio/interop/cats.scala
adamgfraser: It seems like there is enough flexibility in the laws that this isn't an issue but consider:

```scala
ZIO.uninterruptibleMask(_ => ZIO.interrupt *> ZIO.effectTotal(println("I'm here"))
```

Versus the equivalent in Cats Effect 3:

```scala
F.uncancelable(_ => F.canceled *> F.delay(println("I'm here"))
```

In ZIO, "I'm here" will never be printed because while the uninterruptible region prevents the fiber from being interrupted externally `ZIO.interrupt` causes the fiber to die with an interruption cause, which in turn could be recovered from using operators like `run`.

In Cats Effect 3, in contrast, what they call "self-cancelation" is suppressed in an uncancelable region, so "I'm here" will be printed, but then the fiber would be subject to cancelation as soon as it completed the uncancelable region.

You can see this in the difference of the signature of `interrupt`, which returns a `UIO[Nothing`, and `canceled`, which must return a `F[Unit]` because it may actually return successfully if it is in an uninterruptible region.

It seems like we are satisfying the laws with the latest changes so this may not be an issue for this PR in particular but it is a pretty significant difference in the interruption model with a lot of other implications.

interop-cats/shared/src/test/scala/zio/interop/catzSpecBase.scala
joroKr21: Damn, I was thinking about that at some point and then I forgot 👍

interop-cats/jvm/src/main/scala/zio/interop/ZioAsync.scala
joroKr21: Awesome 👍

interop-cats/shared/src/test/scala/zio/interop/catzSpecBase.scala
joroKr21: Ah I remember what was the difficulty here - we need version specific sources 🙈

interop-cats/shared/src/main/scala/zio/interop/cats.scala
joroKr21: Hmm ok I see, thanks for the explanation - I hope this difference wouldn't be a problem in practice.

joroKr21:
@adamgfraser only the fs2 tests are left now - but that looks very strange, I have no clue why they hang 🤔

adamgfraser:
@joroKr21 It looks like one of them relates to how `die` failures are handled. The finalizer in `bracket` is never being invoked but if you replace `ZIO.die` with `ZIO.fail` there the test passes. I'm not sure about the other one.

interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -35,348 +38,444 @@ abstract class CatsEffectPlatform
...
+    ZIO.yieldNow
+
+  override final def forceR[A, B](fa: F[A])(fb: F[B]): F[B] =
+    fa.either *> fb
+
+  override final def uncancelable[A](body: Poll[F] => F[A]): F[A] =
+    ZIO.uninterruptibleMask(body.compose(toPoll))
+
+  override final val canceled: F[Unit] =
+    ZIO.interrupt.unit
```
adamgfraser, line 204:
```suggestion
    ZIO.interrupt
```
Don't need `unit` here due to variance.

interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -35,348 +38,444 @@ abstract class CatsEffectPlatform
...
+  private def trySet(a: A): F[Boolean] =
+    set(a).as(true)
 
-    (run(fa) raceWith run(fb))(
-      { case (l, f) => l.fold(f.interrupt *> ZIO.halt(_), ZIO.succeedNow).map(lv => Left((lv, toFiber(f)))) },
-      { case (r, f) => r.fold(f.interrupt *> ZIO.halt(_), ZIO.succeedNow).map(rv => Right((toFiber(f), rv))) }
-    ).resetForkScope
-  }
+  override def access: F[(A, A => F[Boolean])] =
+    get.map(_ -> trySet)
```
adamgfraser, line 254: I think we need a more complicated state machine here because there are a couple of guarantees we have to honor. First, the `setter` is only supposed to set the value if the current value is still equal to the snapshot value returned by `access`. Second, the `setter` is supposed to only support being called once and return `false` on subsequent invocation. Something like:

```scala
  override def access: F[(A, A => F[Boolean])] =
    get.map { currentValue =>
      val called = new AtomicBoolean(false)
      def setter(a: A): F[Boolean] =
        ZIO.effectSuspendTotal {
          if (called.getAndSet(true)) {
            ZIO.succeedNow(false)
          } else {
            ref.modify { updatedValue =>
              if (currentValue == updatedValue) (true, a)
              else (false, updatedValue)
            }
          }
        }
      (currentValue, setter)
    }
```

interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -35,348 +38,444 @@ abstract class CatsEffectPlatform
...
+    fa
+
+  override final def map2[A, B, Z](fa: F[A], fb: F[B])(f: (A, B) => Z): F[Z] =
+    fa.zipWith(fb)(f)
+
+  override final def as[A, B](fa: F[A], b: B): F[B] =
+    fa.as(b)
+
+  override final def whenA[A](cond: Boolean)(f: => F[A]): F[Unit] =
+    ZIO.effectSuspendTotal(f).when(cond)
```
adamgfraser, line 318:
```suggestion
    ZIO.when(cone)(f)
```

interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -35,348 +38,444 @@ abstract class CatsEffectPlatform
...
+  override def ref[A](a: A): F[effect.Ref[F, A]] =
+    ZRef.make(a).map(new ZioRef(_))
+
+  override def deferred[A]: F[Deferred[F, A]] =
+    Promise.make[E, A].map(new ZioDeferred(_))
 
-  override final def toIO[A](fa: RIO[R, A]): effect.IO[A] =
-    effect.ConcurrentEffect.toIOFromRunCancelable(fa)(this)
+  override final def start[A](fa: F[A]): F[effect.Fiber[F, E, A]] =
+    fa.interruptible.fork.map(toFiber)
```
adamgfraser, line 189:
```suggestion
    fa.interruptible.forkDaemon.map(toFiber)
```
Fibers need to be forked as daemon fibers or auto-supervision can cause fibers to be interrupted in a way that is not possible to manage to manage from Cats Effect 3 code due. With this change the `parJoin` test passes.

interop-cats/shared/src/main/scala/zio/interop/cats.scala
joroKr21: Good point - I guess the same applies to `tryUpdate`, `tryModify` and `tryModifyState`

interop-cats/shared/src/main/scala/zio/interop/cats.scala
joroKr21: Ahh good catch 👍

joroKr21:
> @joroKr21 It looks like one of them relates to how die failures are handled. The finalizer in bracket is never being invoked but if you replace ZIO.die with ZIO.fail there the test passes. I'm not sure about the other one.

Perhaps this test is invalid? Consider the definition of `toOutcome`:
```scala
@inline private[interop] def toOutcome[R, E, A](exit: Exit[E, A]): Outcome[ZIO[R, E, *], E, A] =
    exit match {
      case Exit.Success(value) =>
        Outcome.Succeeded(ZIO.succeed(value))
      case Exit.Failure(cause) if cause.interrupted =>
        Outcome.Canceled()
      case Exit.Failure(cause) =>
        cause.failureOrCause match {
          case Left(error)  => Outcome.Errored(error)
          case Right(cause) => Outcome.Succeeded(ZIO.halt(cause))
        }
    }
```

It's not possible to convert `Right(cause)` to `Outcome.Errored`

interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -35,348 +38,444 @@ abstract class CatsEffectPlatform
...
+  override final def forceR[A, B](fa: F[A])(fb: F[B]): F[B] =
+    fa.either *> fb
+
+  override final def uncancelable[A](body: Poll[F] => F[A]): F[A] =
+    ZIO.uninterruptibleMask(body.compose(toPoll))
+
+  override final val canceled: F[Unit] =
+    ZIO.interrupt.unit
+
+  override final def onCancel[A](fa: F[A], fin: F[Unit]): F[A] =
```
adamgfraser, line 206: Somewhat counterintuitively, we need to run the finalizer here if `fa` dies or is interrupted. In Cats Effect 3 `bracket` is implemented in terms of `uncancelable`, `onCancel`, and `onError`. If an effect dies `onError` can't run because there is no `E` value` but it is clearly not a success. So if `onInterrupted` doesn't run the finalizer then we can leak finalizers in a die failure. While we can override the default implementation of `bracket` users are entitled to rely on the default implementation being correct and rely on that behavior in creating their own resource handling logic as FS2, which is why that test was failing.

adamgfraser:
@joroKr21 Submitted another pull request to your branch. With these changes all tests are passing.

interop-cats/shared/src/main/scala/zio/interop/cats.scala
adamgfraser: I think those are fine  because `update` and `modify` are atomic so they will always set the value based on a consistent snapshot of the initial value. We're just retrying under the hood until we succeed because we don't implement the very low level atomic operations at the effect system level. It is only in `access` where we are promising the ability to potentially do a set based on a consistent view of the state outside `modify`.

interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -35,348 +39,462 @@ abstract class CatsEffectPlatform
...
+      (exit, fiber) => ZIO.succeedNow(Right((toFiber(fiber), toOutcome(exit))))
+    )
+
+  override final def both[A, B](fa: F[A], fb: F[B]): F[(A, B)] =
+    fa zipPar fb
+
+  override final def guarantee[A](fa: F[A], fin: F[Unit]): F[A] =
+    fa.ensuring(fin.orDieWith(fiberFailure))
+
+  override final def bracket[A, B](acquire: F[A])(use: A => F[B])(release: A => F[Unit]): F[B] =
```
adamgfraser, line 232: I think we probably need to suppress the failure here and in `guarantee` to be consistent with `onCancel`.

joroKr21: Makes sense 👍

interop-cats/shared/src/main/scala/zio/interop/cats.scala
joroKr21: Yes, that's because we can't convert `Exit` with die to `Outcome.Errored`.
Pretty clever solution 👍 (I think it's fine, even if it's not exactly `onCancel`).

joroKr21:
> [info] ZIO Test
[info] - summons from catz.core work with only a cats-core dependency
[info]   - ZIO instances
[info]   - ZManaged instances

That works locally for me 🤔

joroKr21:
@adamgfraser success 🥳 🕺 

idk how we want to manage CE2 vs CE3 - maybe a separate branch?

adamgfraser:
@joroKr21 Fantastic! Yes I was thinking we could merge this into `master` and create a separate `ce2` branch.

@ghostdogpr @neko-kai What do you think?

adamgfraser:
@joroKr21 Is there anything else outstanding on this from your perspective?

joroKr21:
> @joroKr21 Is there anything else outstanding on this from your perspective?

Not as far as I can tell. The only awkward part is using `Dispatcher` but I couldn't find a way to make it more ergonomic.
Perhaps worth another try from someone else.

adamgfraser:
@joroKr21 Okay. I will go through everything one more time but I think this is ready for review by others.

interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -35,348 +39,453 @@ abstract class CatsEffectPlatform
...
+  override final def start[A](fa: F[A]): F[effect.Fiber[F, E, A]] =
+    fa.interruptible.forkDaemon.map(toFiber)
+
+  override def never[A]: F[A] =
+    ZIO.never
+
+  override final def cede: F[Unit] =
+    ZIO.yieldNow
+
+  override final def forceR[A, B](fa: F[A])(fb: F[B]): F[B] =
```
adamgfraser, line 195: I think if you change the implementation to recover from all errors exception interruption (including recovering from `Die` failures) then we can `Die` if a finalizer fails in `onCancel`, `guarantee`, or `bracket`.

joroKr21: Interesting - do you have something like this in mind? 
```scala
    fa.run.flatMap {
      case Exit.Success(_)                          => fb
      case Exit.Failure(cause) if cause.interrupted => ZIO.halt(cause)
      case Exit.Failure(_)                          => fb
    }
```

joroKr21: It doesn't work with `onCancel` - the old problem resurfaces

adamgfraser: Really? All the tests passed for me locally?

I used:

```scala
  override final def forceR[A, B](fa: F[A])(fb: F[B]): F[B] =
    fa.foldCauseM(
      cause => if (cause.interrupted) ZIO.halt(cause) else fb,
      _ => fb
    )

  override final def onCancel[A](fa: F[A], fin: F[Unit]): F[A] =
    fa.onError(cause => fin.orDieWith(fiberFailure).unless(cause.failed))

  override final def guarantee[A](fa: F[A], fin: F[Unit]): F[A] =
    fa.ensuring(fin.orDieWith(fiberFailure))

  override final def bracket[A, B](acquire: F[A])(use: A => F[B])(release: A => F[Unit]): F[B] =
    acquire.bracket(release.andThen(_.orDieWith(fiberFailure)), use)
```

joroKr21: Ok I will try again later - maybe they timed out because I was running something else in the background

interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -35,348 +39,453 @@ abstract class CatsEffectPlatform
...
+    fa.memoize
+
+  override final def racePair[A, B](fa: F[A], fb: F[B]) =
+    (fa.interruptible raceWith fb.interruptible)(
+      (exit, fiber) => ZIO.succeedNow(Left((toOutcome(exit), toFiber(fiber)))),
+      (exit, fiber) => ZIO.succeedNow(Right((toFiber(fiber), toOutcome(exit))))
+    )
+
+  override final def both[A, B](fa: F[A], fb: F[B]): F[(A, B)] =
+    fa zipPar fb
```
adamgfraser, line 217:
```suggestion
    fa.interruptible zipPar fb.interruptible
```

interop-cats/shared/src/main/scala/zio/interop/cats.scala
adamgfraser: The timeout for the tests is extremely short at 2 seconds. By comparison in ZIO we use 60 seconds. I noticed one test `ZIO interruption is tied to F interruption` failing on CI but wasn't able to replicate it locally with `nonFlaky(100000)` and a longer timeout.

interop-cats/shared/src/main/scala/zio/interop/CHub.scala
```diff
@@ -149,51 +150,51 @@ object CHub {
    *
    * For best performance use capacities that are powers of two.
    */
-  def bounded[F[+_]: Effect, A](requestedCapacity: Int)(implicit runtime: Runtime[Any]): F[Hub[F, A]] =
-    toEffect(ZHub.bounded[A](requestedCapacity).map(hub => CHub(hub)))
+  def bounded[F[+_]: Async: Dispatcher, A](requestedCapacity: Int)(implicit runtime: Runtime[Any]): F[Hub[F, A]] =
+    ZHub.bounded[A](requestedCapacity).map(hub => CHub(hub)).toEffect[F]
```
adamgfraser, line 154: The type parameter in `toEffect` here and below can be elided, right?

interop-cats/jvm/src/test/scala/zio/interop/Fs2ZioSpec.scala
```diff
@@ -40,23 +36,17 @@ object Fs2ZioSpec extends DefaultRunnableSpec {
...
       released <- Promise.make[Nothing, Unit]
       fail     <- Promise.make[Nothing, Unit]
       _ <- Stream
             .bracket(started.succeed(()).unit)(_ => released.succeed(()).unit)
-            .evalMap[Task, Unit] { _ =>
-              fail.await *> IO.fail(new Exception())
-            }
-            .compile
+            .evalMap[Task, Unit](_ => fail.await *> IO.fail(new Exception()))
+            .compile[Task, Task, Unit]
```
adamgfraser, line 47: Type parameters not necessary here?

interop-cats/jvm/src/test/scala/zio/interop/Fs2ZioSpec.scala
```diff
@@ -22,7 +18,7 @@ object Fs2ZioSpec extends DefaultRunnableSpec {
           }
         },
         testM("works if F is zio.interop.Task") {
-          testCaseJoin[zio.Task].map { ints =>
+          testCaseJoin[zio.RIO[ZEnv, *]].map { ints =>
```
adamgfraser, line 21: Can we provide an instance for `Task` that uses the live services?

interop-cats/shared/src/main/scala/zio/interop/CQueue.scala
```diff
@@ -29,64 +30,76 @@ final class CQueue[F[+_], -A, +B] private[interop] (
   /**
    * @see [[ZQueue.awaitShutdown]]
    */
-  def awaitShutdown(implicit R: Runtime[Any], F: LiftIO[F]): F[Unit] = toEffect(underlying.awaitShutdown)
+  def awaitShutdown(implicit R: Runtime[Any], F: Async[F]): F[Unit] =
```
adamgfraser, line 33: Could be handled in a follow up but would it be better here to require the capabilities for the creation of the queue and not for each operator? That seems more idiomatic to me from Cats Effect 3 code I have seen.

interop-cats/shared/src/main/scala/zio/interop/Schedule.scala
```diff
@@ -108,20 +109,20 @@ final class Schedule[F[+_], -In, +Out] private (private[Schedule] val underlying
   /**
    * @see zio.ZSchedule.addDelayM
    */
-  def addDelayM(f: Out => F[Duration])(implicit R: Runtime[Any], F: Effect[F]): Schedule[F, In, Out] =
+  def addDelayM(f: Out => F[Duration])(implicit F: Dispatcher[F]): Schedule[F, In, Out] =
```
adamgfraser, line 112: Could also push capability constraints into schedule constructors.

interop-cats/shared/src/main/scala/zio/interop/CQueue.scala
kubukoz: Confirmed :)

kubukoz: Definitely if `Async` is involved - you want to minimize the # of places where it's exposed.

adamgfraser: 😃

interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -35,348 +39,453 @@ abstract class CatsEffectPlatform
...
 
-  final override def pure[A](x: A): ParIO[R, E, A] =
-    Par(ZIO.succeedNow(x))
+  final override def pure[A](x: A): F[A] =
+    ParallelF[G, A](ZIO.succeedNow(x))
 
-  final override def map2[A, B, Z](fa: ParIO[R, E, A], fb: ParIO[R, E, B])(f: (A, B) => Z): ParIO[R, E, Z] =
-    Par(Par.unwrap(fa).interruptible.zipWithPar(Par.unwrap(fb).interruptible)(f))
+  final override def map2[A, B, Z](fa: F[A], fb: F[B])(f: (A, B) => Z): F[Z] =
+    ParallelF(ParallelF.value(fa).zipWithPar(ParallelF.value(fb))(f))
```
adamgfraser, line 397:
```suggestion
    ParallelF(ParallelF.value(fa.interruptible).zipWithPar(ParallelF.value(fb.interruptible))(f))
```

interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -35,348 +39,453 @@ abstract class CatsEffectPlatform
...
 
-  final override def ap[A, B](ff: ParIO[R, E, A => B])(fa: ParIO[R, E, A]): ParIO[R, E, B] =
-    Par(Par.unwrap(ff).interruptible.zipWithPar(Par.unwrap(fa).interruptible)(_(_)))
+  final override def ap[A, B](ff: F[A => B])(fa: F[A]): F[B] =
+    map2(ff, fa)(_ apply _)
 
-  final override def product[A, B](fa: ParIO[R, E, A], fb: ParIO[R, E, B]): ParIO[R, E, (A, B)] =
-    Par(Par.unwrap(fa).interruptible.zipPar(Par.unwrap(fb).interruptible))
+  final override def product[A, B](fa: F[A], fb: F[B]): F[(A, B)] =
+    ParallelF(ParallelF.value(fa).zipPar(ParallelF.value(fb)))
```
adamgfraser, line 403:
```suggestion
    ParallelF(ParallelF.value(fa.interruptible).zipPar(ParallelF.value(fb.interruptible)))
```

interop-cats/shared/src/main/scala/zio/interop/CHub.scala
joroKr21: No, looks like type inference doesn't quite manage

adamgfraser: Really? Which version?

interop-cats/jvm/src/test/scala/zio/interop/Fs2ZioSpec.scala
joroKr21: I can give it a try but I suspect implicit priorities could be messed up due to contravariance

interop-cats/shared/src/main/scala/zio/interop/CQueue.scala
joroKr21: Good point, in this case I mindlessly replaced the constraints to make it compile 😄

interop-cats/shared/src/main/scala/zio/stream/interop/FS2StreamSyntax.scala
```diff
@@ -20,7 +20,7 @@ class ZStreamSyntax[R, E, A](private val stream: ZStream[R, E, A]) extends AnyVa
 
   /** Convert a [[zio.stream.ZStream]] into an [[fs2.Stream]]. */
   def toFs2Stream: fs2.Stream[ZIO[R, E, *], A] =
-    fs2.Stream.resource(stream.process.toResourceZIO).flatMap { pull =>
+    fs2.Stream.resource[ZIO[R, E, *], ZIO[R, Option[E], Chunk[A]]](stream.process.toResourceZIO).flatMap { pull =>
```
adamgfraser, line 23: Another Scala 3 issue?

interop-cats/shared/src/main/scala/zio/interop/CHub.scala
joroKr21: Scala 2.13 - I wonder if it matters that `F[+_]` is covariant - we don't actually need it here.
Maybe this will go away if we take `Async` and `Dispatcher` as constructor parameters.

interop-cats/shared/src/main/scala/zio/stream/interop/FS2StreamSyntax.scala
joroKr21: Yes 👍

adamgfraser: I hope we are capturing these somewhere.

interop-cats/shared/src/main/scala/zio/interop/CHub.scala
adamgfraser: Hmmm, this compiled for me on `2.13.5`:

```scala
  def bounded[F[+_]: Async: Dispatcher, A](requestedCapacity: Int)(implicit runtime: Runtime[Any]): F[Hub[F, A]] =
    ZHub.bounded[A](requestedCapacity).map(hub => CHub(hub)).toEffect
```

joroKr21: Ahh perhaps - what didn't compile for me was the helper method called directly:
`toEffect(ZHub.bounded[A](requestedCapacity).map(hub => CHub(hub)))`

adamgfraser: This compiles for me as well:

```scala
  def bounded[F[+_]: Async: Dispatcher, A](requestedCapacity: Int)(implicit runtime: Runtime[Any]): F[Hub[F, A]] =
    toEffect(ZHub.bounded[A](requestedCapacity).map(hub => CHub(hub)))
```

joroKr21: Ah they don't compile if we import `import zio.interop.catz._` instead of only `import zio.interop.catz.zManagedSyntax`

interop-cats/shared/src/main/scala/zio/interop/CQueue.scala
joroKr21: @adamgfraser there are some deprecated methods on `ZQueue` - what should we do with them?

interop-cats/shared/src/main/scala/zio/stream/interop/FS2StreamSyntax.scala
joroKr21: The problem is that we have `Async[RIO[Clock & Blocking, *]]` and `GenConcurrent[ZIO[R, E, *], E]` and here we need `MonadCancel[F, _]`. `Async` takes priority because it's a subtype. But somehow Scala 2 does a better job and picks the `Concurrent` instance. `compile` has the same problem.

joroKr21: I think the expected result type `fs2.Stream[ZIO[R, E, *], A]` should be taken into account by type inference.

interop-cats/shared/src/main/scala/zio/interop/CHub.scala
joroKr21: 2.12 also has a problem: https://github.com/zio/interop-cats/pull/263/checks?check_run_id=2388227935

interop-cats/jvm/src/test/scala/zio/interop/Fs2ZioSpec.scala
joroKr21: I added instances that depend on the runtime: 28b4d37

interop-cats/shared/src/main/scala/zio/interop/CQueue.scala
adamgfraser: @joroKr21 We should mark them as deprecated for now and then remove them when we release a version that supports ZIO 2.0, which is I think what we are currently doing.

interop-cats/shared/src/main/scala/zio/interop/CHub.scala
adamgfraser: Oh I see. Because the `F` in `toEffect` being `ZIO[R, E, *]` is a valid solution as well.

interop-cats/jvm/src/test/scala/zio/interop/Fs2ZioSpec.scala
adamgfraser: Nice! Any implicit prioritization issues with defining those instances at the same level as the more polymorphic ones?

interop-cats/shared/src/main/scala/zio/interop/CQueue.scala
adamgfraser: @joroKr21 Looks much nicer with your changes now!

adamgfraser (approved):
Looks good to me!

interop-cats/jvm/src/test/scala/zio/interop/Fs2ZioSpec.scala
joroKr21: Not really. Note that those instances fix the environment to `Any`. Alternatively we could try to leave it generic, but I'm not convinced - usually you have either no environment or you have at least `ZEnv` and add more on top. WDYT?

adamgfraser: I think it is fine.

adamgfraser:
@joroKr21 Thank you for all your hard work on this! Users are going to really appreciate having this functionality.

kubukoz:
🎉

Fristi:
🎉 Lots of goodness! Thanks!

agilesteel:
Awesome work! An update to the readme would be much appreciated. I can't seem to be able to get basic things to compile since the API changed a bit.

In particular I find it challenging to find instances for `cats.effect.std.Console[zio.Task]` or `zio.Task.concurrentEffectWith` does not seem to exist any longer.

Thanks!

kubukoz:
There is no ConcurrentEffect.

kubukoz:
And I think there is no Console instance yet - probably can just be added.

For the CE side of things, here's the migration guide https://typelevel.org/cats-effect/docs/migration-guide

agilesteel:
I figured it out eventually.

```scala
import cats.effect.std
import zio.*
import zio.interop.catz.*

given std.Console[RIO[ZEnv, *]] =
  std.Console.make
```

adamgfraser:
Yes that would be great.

@joroKr21 Do you want to work on this?

joroKr21:
Is `Console` something new in CE3?

joroKr21:
> @joroKr21 Do you want to work on this?

It should be quite easy, if nobody beats me to it.

calvinlfer:
> > @joroKr21 Do you want to work on this?
> 
> It should be quite easy, if nobody beats me to it.

Done 😸

PR #275, opened 2020-12-28
Type class instances for Chunk and NonEmptyChunk

guymers:
Type class instances for Chunk and NonEmptyChunk are now defined for the
same type classes that Cats provides for Vector and NonEmptyVector
respectively.


neko-kai (approved):

neko-kai:
Thank you! 🙏

PR #288, opened 2021-02-09
Expose R parameter for ioTimer

blast-hardcheese:
I don't know if we care about maintaining backwards compatibility for the `ioTimer` function, I can amend this PR to remove it, but I'm currently having to `asInstanceOf` in my own code because `Timer` is invariant

### DIFF
--- a/interop-cats/shared/src/main/scala/zio/interop/cats.scala
+++ b/interop-cats/shared/src/main/scala/zio/interop/cats.scala
@@ -46,7 +46,7 @@ abstract class CatsEffectPlatform
   }
 
   object implicits {
-    implicit final def ioTimer[E]: effect.Timer[IO[E, *]] = ioTimer0.asInstanceOf[effect.Timer[IO[E, *]]]
+    implicit final def ioTimer[R, E]: effect.Timer[ZIO[R, E, *]] = ioTimer0.asInstanceOf[effect.Timer[ZIO[R, E, *]]]
 
     private[this] val ioTimer0: effect.Timer[IO[Any, *]] =
       zioClock.toTimer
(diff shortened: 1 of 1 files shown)


neko-kai:
@blast-hardcheese `rioTimer` has the same erasure as `ioTimer`, so you may just replace `ioTimer` - this will not affect bincompat.

interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -46,7 +46,8 @@ abstract class CatsEffectPlatform
   }
 
   object implicits {
-    implicit final def ioTimer[E]: effect.Timer[IO[E, *]] = ioTimer0.asInstanceOf[effect.Timer[IO[E, *]]]
+    final def ioTimer[E]: effect.Timer[IO[E, *]]                  = ioTimer0.asInstanceOf[effect.Timer[IO[E, *]]] // bincompat?
+    implicit final def rioTimer[R, E]: effect.Timer[ZIO[R, E, *]] = ioTimer0.asInstanceOf[effect.Timer[ZIO[R, E, *]]]
 
     private[this] val ioTimer0: effect.Timer[IO[Any, *]] =
```
neko-kai, line 52: Might change this signature to 

```suggestion
    private[this] val ioTimer0: effect.Timer[ZIO[Any, Nothing, *]] =
```

To make it more understandable that its type-safe to cast, since `ZIO[Any, Nothing, ?] <: ZIO[R, E, ?] ∀ R, E`

blast-hardcheese:
@neko-kai Resolved

neko-kai (approved):

neko-kai:
@blast-hardcheese Thank you! 🙏

PR #314, opened 2021-03-12
Fix: runAsync

### LINKED ISSUE #313: Issue with fs2-blobstore and ZIO
fs2-blobstore is a library which has a `Store` abstraction for storing blobs. The nice thing is that you can easily swap an implementation, this is useful during testing (local) and migrating to other blob stores.

We use it in a app where we also use ZIO and since the interop-cats module I would not expect any problems, yet it seems there is an issue which seems to be ZIO specific.

While downloading a file, from time to time it seems the download is incomplete. After checking the logs I found out:

> java.lang.Error: received record [java.nio.HeapByteBuffer[pos=0 lim=2875 cap=2875]] in invalid state [UpstreamCompletion]

It seems that the FSM being used in the fs2-reactivestreams interop is already completed (state: UpstreamCompletion), while another chunk of bytes is still sent in that state causing that error.

I already verified that this is not an issue in the fs2-blobstore implementation by writing a property test: https://github.com/fs2-blobstore/fs2-blobstore/pull/334

I've created a reproduction case in a repository here: https://github.com/vectos/zio-fs2-blobstore-issue/

Fristi:
Locally it seems to resolve #313, bit confused here. The SyncIO seems values returned seem to use Async variants, shouldn't this be the sync variants as SyncIO implies?

### DIFF
--- a/interop-cats/shared/src/main/scala/zio/interop/cats.scala
+++ b/interop-cats/shared/src/main/scala/zio/interop/cats.scala
@@ -173,8 +173,9 @@ private class CatsConcurrentEffect[R](rts: Runtime[R])
     cb: Either[Throwable, A] => effect.IO[Unit]
   ): effect.SyncIO[Unit] =
     effect.SyncIO {
-      rts.unsafeRunAsync(fa.run) { exit =>
-        cb(exit.flatMap(identity).toEither).unsafeRunAsync(_ => ())
+      rts.unsafeRunSync(fa) match {
+        case Exit.Success(value) => cb(Right(value)).unsafeRunSync()
+        case Exit.Failure(cause) => cb(Left(cause.squash)).unsafeRunSync()
       }
     }
 


neko-kai (changes requested):
This is not a viable solution since it turns runAsync into runSync (the argument effect is being run all the way to completion on the current thread)

I wonder if there's another way to fix that issue.

Regarding the confusion - SyncIO returned by runAsync is supposed to be an effect that initiates the asynchronous execution of the given effect and sets up the given callback. However, it will return before the given effect is completed and the way it initiates the async execution shouldn't use any async capabilities and should be "safe" to execute using "unsafeRunSync" (that is - when ran with unsafeRunSync it must execute to completion and not contain any Async nodes - so that it doesn't require multi-threading that may be unavailable on JS for example)

Fristi:
Thank you for your explanation @neko-kai.

I've digged some more and found out that the fs2-reactiive-streams dependency was outdated. See #313. By updating it's fixed and there was nothing wrong here.

PR #344, opened 2021-04-27
Update Scala 3 to 3.0.0-RC3

Lermex:


adamgfraser (changes requested):
@Lermex Thanks for working on this! Looks like there is an issue with one of the dependencies. Good to go once that is resolved!

adamgfraser (approved):

PR #353, opened 2021-05-10
Improve ergonomics (opt-in) at the cost of testability

calvinlfer:
- Provide easier import syntax to avoid explicitly bringing in the runtime
- Add documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,11 @@
 
 This library provides instances required by Cats Effect.
 
+## Installation
+```sbt
+libraryDependencies += "dev.zio" %% "zio-interop-cats" % "<check-badge-for-latest-version>"
+```
+
 ## `ZIO` Cats Effect 3 instances
 
 **ZIO** integrates with Typelevel libraries by providing an instance of `Concurrent`, `Temporal` and `Async` for `Task` 
@@ -14,7 +19,7 @@ as required, for instance, by `fs2`, `doobie` and `http4s`.
 For convenience, we have defined an alias as follows:
 
 ```scala
-  type Task[A] = ZIO[Any, Throwable, A]
+type Task[A] = ZIO[Any, Throwable, A]
 ```
 
 Therefore, we provide Cats Effect instances based on this specific datatype.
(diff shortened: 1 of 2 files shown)


neko-kai:
@calvinlfer 
Hi, in CE2 version we've used a `zio.interop.catz.implicits` object for implicits that added convenience but broke testability, so I think you may be able to return that object and put the Runtime implicit there.

calvinlfer:
> @calvinlfer
> Hi, in CE2 version we've used a `zio.interop.catz.implicits` object for implicits that added convenience but broke testability, so I think you may be able to return that object and put the Runtime implicit there.

I will do this, thanks @neko-kai

calvinlfer:
Addressed as per @neko-kai's recommendation 😸

interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -46,6 +46,21 @@ abstract class CatsEffectPlatform
...
+   * summon Cats Effect typeclasses without the ceremony of
+   *
+   * {{{
+   * ZIO.runtime[Clock with Blocking].flatMap { implicit runtime =>
+   *  implicit val asyncTask: Async[Task] = implicitly
+   *  ...
+   * }
+   * }}}
+   */
+  object implicits {
```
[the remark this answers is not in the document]
calvinlfer: Fixed!

README.md
```diff
@@ -69,14 +74,51 @@ def ceAsync =
...
+    .awakeDelay[Task](10.seconds) // this type annotation is mandatory
+    .evalTap(in => cats.effect.std.Console.make[Task].println(s"Hello $in"))
+    .compile
+    .drain
+```
+
+Rather than requiring the runtime implicit, we can add an import (if we don't mind depending on `Runtime.default`):
+```scala
+import zio.interop.catz._
+import zio.interop.implicits._
```
neko-kai, line 102: Sorry! just this one place needs to be updated and that'll be it 🙏

calvinlfer: oops this is my bad! i will fix now sorry for the trouble!

PR #367, opened 2021-05-28
Include Chunk Instances In Default Scope

### LINKED ISSUE #366: Consider Providing Cats Core Instances More Directly
Right now if you want to get Cats instances for some ZIO data types you need to do `import zio.interop.catz.core._` instead of `import zio.interop.catz._`. For example the `Traverse` instance for `Chunk` is not in scope without that import. Should we have `catz` directly extend these instances so the user can get everything they need in one import? Otherwise I worry it can be somewhat confusing for the user, especially since some Cats instances such as `Monad` for `UIO` are in scope without doing this because they are implemented by Cats Effect instances.

@neko-kai @ghostdogpr What do you think?

adamgfraser:
Resolves #366.


neko-kai (approved):
👍

PR #374, opened 2021-06-28
RFC:  using % Provided

sideeffffect:
What would you guys think of making the dependencies on both zio-* and cats-* libraries as ` % Provided`?
I assume folks would bring in these libraries explicitly and want `interop-cats` only to bring the type classes for the libraries they actually use.
I learned that `interop-cats` depends firmly on all the zio-* and cats-* libs when I realized I have a zio-test JAR in my application image.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -62,7 +62,7 @@ lazy val interopCats = crossProject(JSPlatform, JVMPlatform)
         "org.typelevel" %%% "cats-mtl"        % catsMtlVersion,
         "co.fs2"        %%% "fs2-core"        % fs2Version
       )
-      ("dev.zio" %%% "zio" % zioVersion) :: (if (scalaVersion.value.beginsWith("3")) optLibraries.map(_ % Optional) else optLibraries)
+      ("dev.zio" %%% "zio" % zioVersion) :: (if (scalaVersion.value.startsWith("3")) optLibraries.map(_ % Optional) else optLibraries)
     },
     libraryDependencies ++= Seq(
       "dev.zio"       %%% "zio-test-sbt"         % zioVersion,
--- a/build.sbt
+++ b/build.sbt
@@ -55,7 +55,7 @@ lazy val interopCats = crossProject(JSPlatform, JVMPlatform)
   .settings(buildInfoSettings)
   .settings(
     libraryDependencies ++= Seq(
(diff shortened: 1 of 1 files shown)


joroKr21:
Oh damn I could swear they used to be `Optional` but we had to change that because of Scala 3 and I think we forgot to look into it afterwards 🙈

neko-kai:
@joroKr21 The non-optionality was only restricted to Scala 3 at first, I missed the change when it became global...
@sideeffffect Could you please make the `% Optional` qualifier dependent on Scala version as in
```scala
libraryDependencies ++= {
  val optLibraries = Seq(...)
  if (scalaVersion.value.beginsWith("3")) optLibraries.map(_ % Optional) else optLibraries
}
```

build.sbt
```diff
@@ -54,14 +54,17 @@ lazy val interopCats = crossProject(JSPlatform, JVMPlatform)
...
-      "dev.zio"       %%% "zio"             % zioVersion,
-      "dev.zio"       %%% "zio-streams"     % zioVersion        % Provided,
-      "dev.zio"       %%% "zio-test"        % zioVersion        % Provided,
-      "org.typelevel" %%% "cats-effect-std" % catsEffectVersion % Provided,
-      "org.typelevel" %%% "cats-mtl"        % catsMtlVersion    % Provided,
-      "co.fs2"        %%% "fs2-core"        % fs2Version        % Provided
-    ),
+    libraryDependencies ++= {
+      val optLibraries = Seq(
+        "dev.zio"       %%% "zio"             % zioVersion,
```
neko-kai, line 59: Zio is not optional

build.sbt
```diff
@@ -54,14 +54,17 @@ lazy val interopCats = crossProject(JSPlatform, JVMPlatform)
...
+        "dev.zio"       %%% "zio-streams"     % zioVersion        % Provided,
```
neko-kai: Ought to remove Provided from these since Optional replaces them

joroKr21:
> @joroKr21 The non-optionality was only restricted to Scala 3 at first, I missed the change when it became global...

me too

build.sbt
sideeffffect: done :heavy_check_mark:

neko-kai (dismissed):

neko-kai (approved):

sideeffffect:
Thanks :+1:

PR #395, opened 2021-08-27
Make LiftIO instance cancellable

### LINKED ISSUE #388: `LiftIO` instance does not respect cancelation
The current implementation does not use the cancelation token in case of interruption:
https://github.com/zio/interop-cats/blob/429e2270c12125fcc399a86993c5a8465db782b7/interop-cats/shared/src/main/scala/zio/interop/cats.scala#L579-L582

The impl should be rewritten to something like this:
```scala
ZIO.effectAsyncInterrupt { cb =>
  val (fa, canceler) = ioa.unsafeToFutureCancelable()
  cb(ZIO.fromFuture(_ => fa))
  Left(ZIO.fromFuture(_ => canceler()).orDie)
}
```

joroKr21:
Closes #388

### DIFF
--- a/interop-cats/jvm/src/test/scala/zio/interop/CatsInteropSpec.scala
+++ b/interop-cats/jvm/src/test/scala/zio/interop/CatsInteropSpec.scala
@@ -1,19 +1,33 @@
 package zio.interop
 
+import cats.effect.{ IO as CIO, LiftIO }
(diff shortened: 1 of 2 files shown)


adamgfraser (approved):

adamgfraser:
@joroKr21 Thank you! 🙏

PR #418, opened 2021-09-23
Expose toEffect and fromEffect

senia-psm:
Expose `toEffect` and `fromEffect` as public API.

Use case: with `doobie.ConnectionIO` to use `zio` effects in transaction with queries.
```scala
for
  r1 <- selectSomethingForUpdate
  _ <- log.info(s"selected: $r1").toEffect[ConnectionIO]
  r2 <- updateSomething(r1)
yield r2
```

> I agree that and `fromEffect` seem generically useful. 
(c) @adamgfraser

### DIFF
--- a/zio-interop-cats/shared/src/main/scala/zio/interop/package.scala
+++ b/zio-interop-cats/shared/src/main/scala/zio/interop/package.scala
@@ -72,7 +72,7 @@ package object interop {
         }
     }
 
-  @inline private[zio] def fromEffect[F[_], A](fa: F[A])(implicit F: Dispatcher[F]): Task[A] =
+  @inline def fromEffect[F[_], A](fa: F[A])(implicit F: Dispatcher[F]): Task[A] =
     ZIO
       .effectTotal(F.unsafeToFutureCancelable(fa))
       .flatMap { case (future, cancel) =>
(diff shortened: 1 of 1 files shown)


adamgfraser (approved):
This makes sense to me but would appreciate opinions from others.

joroKr21:
I agree those are useful. Maybe worth to ponder the API a bit.
- Do the names `fromEffect` and `toEffect` make sense?
- Does it feel weird to have them as top-level functions?
- We have syntax only in one direction - should we have the other too?

senia-psm:
@joroKr21 
 - In my use case naming makes sense: `zio.toEffect[ConnectionIO]` as "transform ZIO to effect of type ConnectionIO". In general - I'm not aware of better names.
 - Keeping these methods in `interop` instead of `interop.cats` makes sense: instead of manual `ZIO -> cast.effect.IO -> F` it allows devs to skip `cats.effect.IO` step. And adding new methods to `interop.cats` seems dangerous: there are lots of projects with wildcard import from `interop.cats` already.
  - I have no use-cases for `fromEffect`. Adding new method to all `F[_]` might cause some issues with code completion in IDE. We might end up with `fromEffect` method in code completion on all `Option`, `List`, etc.

neko-kai (dismissed):
I don't see why not

adamgfraser:
@senia-psm Can you resolve conflicts and we can get this merged?

senia-psm:
@adamgfraser done

adamgfraser (approved):

adamgfraser:
@senia-psm Thank you! 🙏

PR #440, opened 2021-10-31
Update cats-effect version to 3.2.9

zeal18:
Tests are red because of the IO generator generates failures. It looks like cats-effect-laws uncommented test cases which run IOs and generating failures leads to failing that runs


jdegoes (approved):

jdegoes:
Congratulations on your first pull request merge, and thanks for all your work getting to this point!


giphy-4

PR #451, opened 2021-11-08
Update cats-effect version to 3.2.9

zeal18:
In the latest version cats-effect test framework uncomments additional test cases which run effects. Current zio.IO generator generates failures as well as successes, this is unnatural for ce.IO and the test framework doesn't expect such pre-failed values. Instead of providing pre-failed values the framework generates failures using `MonadError` when it's needed (for example in `fiberErrorIsOutcomeErrored` or `evalOnRaiseErrorIdentity` laws).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -49,7 +49,7 @@ lazy val root = project
 
 val zioVersion                 = "1.0.12"
 val catsVersion                = "2.6.1"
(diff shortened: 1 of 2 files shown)


adamgfraser (approved):
@zeal18 Fantastic! Thank you for working on this! 🙏

PR #455, opened 2021-11-13
Update FS2 and attempt optimizations to FS2 -> ZStream conversion logic

calvinlfer:
Update FS2 integration logic and make optimizations to avoid re-chunking the FS2 stream (by dropping down to fs2's Pull to access the existing chunked structure) & matching on each subtype of the fs2.Chunk to attempt conversion to a proper ZIO counterpart

### DIFF
--- a/zio-interop-cats/shared/src/main/scala/zio/stream/interop/FS2StreamSyntax.scala
+++ b/zio-interop-cats/shared/src/main/scala/zio/stream/interop/FS2StreamSyntax.scala
@@ -2,7 +2,7 @@ package zio
 package stream.interop
 
 import fs2.Stream
-import zio.interop.catz.{ concurrentInstance, zManagedSyntax, zioResourceSyntax }
+import zio.interop.catz.{ concurrentInstance, zManagedSyntax }
 import zio.stream.{ Take, ZStream }
 
 import scala.language.implicitConversions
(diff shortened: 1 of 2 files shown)


zio-interop-cats/shared/src/main/scala/zio/stream/interop/FS2StreamSyntax.scala
```diff
@@ -40,37 +40,65 @@ final class FS2RIOStreamSyntax[R, A](private val stream: Stream[RIO[R, _], A]) {
...
+      fromQueue.drainFork(toQueue)
+    }
+
+  private def toZioChunk(in: fs2.Chunk[A]): Chunk[A] =
+    in match {
+      case fs2.Chunk.ArraySlice(values, _, _) =>
+        Chunk.fromArray(values)
+
+      case buffer: fs2.Chunk.Buffer[_, _, _] =>
+        Chunk.fromIterator(buffer.iterator)
```
calvinlfer, line 93: The documentation for `buffer.iterator` mentions that the returned iterator is not thread-safe so I'm wondering if this is still safe to use?

zio-interop-cats/shared/src/main/scala/zio/stream/interop/FS2StreamSyntax.scala
```diff
@@ -40,37 +40,65 @@ final class FS2RIOStreamSyntax[R, A](private val stream: Stream[RIO[R, _], A]) {
...
+  }
+
+  private def toZStreamSingle[R1 <: R]: ZStream[R1, Throwable, A] = {
+    def integrate(stream: fs2.Stream[RIO[R, _], A], zQueue: Queue[Take[Throwable, A]]): fs2.Pull[RIO[R, _], A, Unit] =
+      stream.pull.uncons.flatMap {
+        case None =>
+          fs2.Pull.eval(zQueue.offer(Take.end)) >> fs2.Pull.done
+
+        case Some((elements, stream)) =>
+          val offer = zQueue.offerAll(toZioChunk(elements).map(Take.single))
```
calvinlfer, line 64: I was not sure whether we could merge this with `Take.chunk(toZioChunk(fs2Chunk))` or is it better to use `toZioChunk(elements).map(Take.single)`?

Also, It might be better to use uncons1 in this single-element case because I feel like I'm building an bounded(chunk-size-of-fs2-stream) atop a single element queue

zio-interop-cats/shared/src/main/scala/zio/stream/interop/FS2StreamSyntax.scala
adamgfraser: Yes this should be fine because there is only one thread interacting with the iterator, the one running `Chunk.fromIterator`, and it will just read the iterator and place its contents into an immutable data structure that will return and will be safe for concurrent access.

zio-interop-cats/shared/src/main/scala/zio/stream/interop/FS2StreamSyntax.scala
adamgfraser: I think you definitely want to offer a single `Take` containing a chunk of elements here.

calvinlfer: Hey @adamgfraser based on this, I can merge both implementations together since the previous chunk counterpart was doing that

calvinlfer:
hey @adamgfraser @neko-kai can y'all please have a look when y'all have some time?

adamgfraser (approved):

PR #460, opened 2021-11-25
Fix tailRecM being not stack safe

### LINKED ISSUE #458: ZIO tailRecM is not stack safe
I'm using `zio` `1.0.12`, `zio-interop-cats` `2.5.1.0`, and `cats-core` `2.6.1`.

---

Consider the following simple implementation of DFS on a `size`-by-`size` grid.

```scala
  private val size = 1000

  private def isIn(coordinates: (Int, Int)): Boolean = {
    val (x, y) = coordinates
    0 <= x && x < size && 0 <= y && y < size
  }

  private def adjacentCoordinates(coordinates: (Int, Int)): List[(Int, Int)] = {
    val (x, y) = coordinates
    List(
      (x - 1, y),
      (x + 1, y),
      (x, y - 1),
      (x, y + 1)
    )
  }

  private def dfsZIO(current: (Int, Int), visited: Set[(Int, Int)]): UIO[Boolean] =
    if (!isIn(current) || visited.contains(current))
      UIO(false)
    else if (current == (0, 0))
      UIO(true)
    else {
      // Not stack safe
      adjacentCoordinates(current).existsM(adjacent => dfsZIO(adjacent, visited + current))
    }
```

Just creating an effect by calling `dfsZIO((size - 1, size - 1), Set())` causes stack overflow.

```
Exception in thread "main" java.lang.StackOverflowError
	at scala.collection.immutable.BitmapIndexedSetNode.mergeTwoKeyValPairs(HashSet.scala:737)
	at scala.collection.immutable.BitmapIndexedSetNode.updated(HashSet.scala:476)
	at scala.collection.immutable.BitmapIndexedSetNode.updated(HashSet.scala:406)
...
```

---

The reason seems to be that `tailRecM`, which is used by `existsM`, is not stack safe.
The monad instance for ZIO is an instance of `CatsMonadError`, which uses `StackSafeMonad` for `tailRecM` implementation.
This is not stack safe because the implementation of `tailRecM` provided by `StackSafeMonad` evaluates `f(a)` eagerly, which in turn causes the recursive part `dfsZIO(adjacent, visited + current)` to be called eagerly as well.

This issue was actually already fixed by https://github.com/zio/zio/pull/795, but that fix seems to be gone now.
I tested this fix in my fork (https://github.com/kjh618/interop-cats/commit/3cb28b7c4af69417f1c0a0c9d71c1d8af387b8e4), and the code above no longer causes stack overflow.
Should I create a PR?

kjh618:
Fixes #458.

Something to note is that the original implementation passes `tailRecMStackSafety` in cats (https://github.com/typelevel/cats/blob/e0784fc0d95b9c4eed0bb1ddce6c69e3fd9a4a4f/laws/src/main/scala/cats/laws/MonadLaws.scala#L39), which does not give `tailRecM` a recursive function.
This PR doesn't change the semantics in that case, but for the cases where `tailRecM` is given a recursive function like in `tailRecMConstructionStackSafety`, it prevents stack overflow during construction.

It's tested on v2.5.1.0, so some changes may be necessary to apply this to master.

### DIFF
--- a/zio-interop-cats/shared/src/main/scala/zio/interop/cats.scala
+++ b/zio-interop-cats/shared/src/main/scala/zio/interop/cats.scala
@@ -291,11 +291,14 @@ private class CatsMonadError[R, E] extends MonadError[ZIO[R, E, *], E] {
 
   override final def attempt[A](fa: ZIO[R, E, A]): ZIO[R, E, Either[E, A]] = fa.either
 
-  override def tailRecM[A, B](a: A)(f: A => ZIO[R, E, Either[A, B]]): ZIO[R, E, B] =
-    flatMap(ZIO.effectSuspendTotal(f(a))) {
-      case Left(a)  => tailRecM(a)(f)
-      case Right(b) => pure(b)
+  override def tailRecM[A, B](a: A)(f: A => ZIO[R, E, Either[A, B]]): ZIO[R, E, B] = {
+    def loop(a: A): ZIO[R, E, B] = f(a).flatMap {
+      case Left(a)  => loop(a)
+      case Right(b) => ZIO.succeedNow(b)
     }
+
+    ZIO.effectSuspendTotal(loop(a))
+  }
 }
 
 /** lossy, throws away errors using the "first success" interpretation of SemigroupK */
--- a/zio-interop-cats/shared/src/main/scala/zio/interop/cats.scala
+++ b/zio-interop-cats/shared/src/main/scala/zio/interop/cats.scala
@@ -272,7 +272,7 @@ private class CatsConcurrent[R] extends CatsMonadError[R, Throwable] with Concur
     ZIO.uninterruptibleMask(_(fa).either.flatMap(f))
 }
 
-private class CatsMonadError[R, E] extends MonadError[ZIO[R, E, *], E] with StackSafeMonad[ZIO[R, E, *]] {
+private class CatsMonadError[R, E] extends MonadError[ZIO[R, E, *], E] {
   override final def pure[A](a: A): ZIO[R, E, A]                                         = ZIO.succeedNow(a)
   override final def map[A, B](fa: ZIO[R, E, A])(f: A => B): ZIO[R, E, B]                = fa.map(f)
   override final def flatMap[A, B](fa: ZIO[R, E, A])(f: A => ZIO[R, E, B]): ZIO[R, E, B] = fa.flatMap(f)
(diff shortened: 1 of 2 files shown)


adamgfraser (approved):
Thank you for your work on this! 🙏

Good to go when conflicts are resolved.

joroKr21:
Looks like there are a lot of unrelated changes - or you based it off some old branch 🤔

kjh618:
Yeah I started my branch from the tag `v2.5.1.0`.

I rebased my branch to `ce2` and changed the PR base branch to `ce2`.

zio-interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -290,6 +290,12 @@ private class CatsMonadError[R, E] extends MonadError[ZIO[R, E, *], E] with Stac
   override final def raiseError[A](e: E): ZIO[R, E, A] = ZIO.fail(e)
 
   override final def attempt[A](fa: ZIO[R, E, A]): ZIO[R, E, Either[E, A]] = fa.either
+
+  override def tailRecM[A, B](a: A)(f: A => ZIO[R, E, Either[A, B]]): ZIO[R, E, B] =
+    flatMap(ZIO.effectSuspendTotal(f(a))) {
+      case Left(a)  => tailRecM(a)(f)
+      case Right(b) => pure(b)
+    }
```
joroKr21, line 298: I just realised that this might be a bit wasteful. Because we call `tailRecM` recursively that means we suspend on every iteration instead of just once. Do you think you could defined a local `loop` function without the suspend and the only suspend once initially? 🤔

joroKr21: Something like this:
```scala
override def tailRecM[A, B](a: A)(f: A => ZIO[R, E, Either[A, B]]): ZIO[R, E, B] = {
    def loop(a: A): ZIO[R, E, B] = f(a).flatMap {
      case Left(a)  => loop(a)
      case Right(b) => ZIO.succeedNow(b)
    }

    ZIO.effectSuspendTotal(loop(a))
  }
```

kjh618: Yep, that sounds good!

joroKr21 (approved):
Nice! This will probably need porting across ZIO / CE versions and also to `Managed` 😄

adamgfraser (approved):

PR #476, opened 2021-12-15
Upgrade ZIO Interop Cats Effect 3 to ZIO 2.0.0-RC1

adamgfraser:


zio-interop-cats-tests/shared/src/test/scala/zio/interop/CatsSpecBase.scala
```diff
@@ -64,9 +62,12 @@ private[zio] trait CatsSpecBase
         case infinite: Infinite     =>
           ZIO.dieMessage(s"Unexpected infinite duration $infinite passed to Ticker")
       }
+
+      def scheduler(implicit trace: ZTraceElement): UIO[Scheduler] =
+        ???
```
ghostdogpr, line 67: Not needed?

ghostdogpr (approved):

zio-interop-cats-tests/shared/src/test/scala/zio/interop/CatsSpecBase.scala
adamgfraser: This is part of the interface of `Clock` in ZIO 2.0 so we need to honor the contract though like some of the other methods above it will never be called in tests so we don't actually have to implement it.

zio-interop-cats/shared/src/main/scala/zio/interop/catsmtl.scala
```diff
@@ -18,31 +18,13 @@ package zio.interop
...
-import zio.{ CanFail, ZIO, ZTraceElement }
-import zio.internal.stacktracer.{ Tracer => CoreTracer }
+import zio.{ CanFail, ZIO }
 import zio.internal.stacktracer.InteropTracer
 
 abstract class CatsMtlPlatform extends CatsMtlInstances
 
 abstract class CatsMtlInstances {
 
-  implicit def zioLocal[R, E](implicit ev: Applicative[ZIO[R, E, _]]): Local[ZIO[R, E, _], R] =
```
lacarvalho91, line 29: @adamgfraser just wondering, why was this and the `Ask` instance removed? can they not be implemented with ZIO 2.x?

adamgfraser: @lacarvalho91 It can be implemented but needs to build in the `ZEnvironment`. See #662.

Issue #509, opened 2022-02-26
Async finalizer not working


durban:
The finalizer action returned from `cats.effect.kernel.Async#async` is never executed. Version 3.3.0-RC2, Cats Effect version 3.3.5.

```scala
package com.example

import scala.concurrent.duration._

import cats.syntax.all._
import cats.effect.kernel.Async

import zio.{ ZIO, Task, Runtime, Clock }
import zio.interop.catz.asyncRuntimeInstance

class Test extends munit.FunSuite {

  def asyncTask[F[_]](implicit F: Async[F]): F[Int] = {
    F.async[Int] { _ =>
      F.pure(Some(F.delay(println("CANCEL"))))
    }
  }

  def task[F[_]](implicit F: Async[F]): F[Unit] = {
    for {
      fib <- F.start(asyncTask[F])
      _ <- F.sleep(1.second)
      _ <- fib.cancel
    } yield ()
  }

  test("Async cancel ZIO") { // does not print "CANCEL"
    val t = ZIO.runtime.flatMap { implicit r: Runtime[Clock] =>
      task[Task]
    }
    zio.Runtime.default.unsafeRunTask(t)
  }

  test("Async cancel CE IO") { // prints "CANCEL"
    val t = task[cats.effect.IO]
    t.unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
  }
}
```

(For comparison, I've added an example with `cats.effect.IO`, which does work as expected.)

neko-kai:
Fixed by #542

PR #542, opened 2022-06-15
Fix #541 #509 respect finalizer in `Async#async` implementation

### LINKED ISSUE #541: Canceler unlawfully ignored in CE3 `Async#async` implementation, can cause hangs etc.
In implementation of `async`, the result of `k` is ignored https://github.com/zio/interop-cats/blob/zio2-interop-ce3/zio-interop-cats/jvm/src/main/scala/zio/interop/ZioAsync.scala#L43= :
```
        k(either => register(promise.await *> ZIO.fromEither(either))) *> promise.succeed(())
```

Going by scaladocs, `k`'s return value is significant:

```
`k` returns an `Option[F[Unit]]` which is an optional finalizer to be run in the event that
the fiber running {{{async(k)}}} is canceled.
```

That is, seems like the the return of `k` ought to be stuffed into `.onInterrupt` and ignoring the finalizer could hang the program if it relies on the finalizer for synchronization in event of interruption for the async, as the description says it should. (Odd that discipline tests missed that)

/cc @joroKr21 @adamgfraser

neko-kai:
/cc @joroKr21

### DIFF
--- a/zio-interop-cats-tests/jvm/src/test/scala/zio/interop/CatsInteropSpec.scala
+++ b/zio-interop-cats-tests/jvm/src/test/scala/zio/interop/CatsInteropSpec.scala
@@ -1,6 +1,6 @@
 package zio.interop
 
-import cats.effect.{ IO as CIO, LiftIO }
+import cats.effect.{ Async, IO as CIO, LiftIO }
 import cats.effect.kernel.{ Concurrent, Resource }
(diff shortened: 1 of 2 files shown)


joroKr21 (approved):
Nice test 👍

adamgfraser (approved):

PR #543, opened 2022-06-15
Fix `toManagedZIO` not propagating restore of outer `uninterruptibleMask` ask` in interpretation of `Resource.Allocated` case

neko-kai:
* Convert `Throwable` Dies to `Outcome.Error` not `Outcome.Succeeded` whenever the error type is `Throwable`
* Remove `.interruptible` from `zio.interop.fromEffect`

/cc @joroKr21

### DIFF
--- a/zio-interop-cats/js/src/main/scala/zio/interop/ZioAsync.scala
+++ b/zio-interop-cats/js/src/main/scala/zio/interop/ZioAsync.scala
@@ -6,7 +6,7 @@ import zio.{ Promise, RIO, ZIO }
 import scala.concurrent.{ ExecutionContext, Future }
 
 private abstract class ZioAsync[R]
-    extends ZioTemporal[R, Throwable]
+    extends ZioTemporal[R, Throwable](Some(_))
     with Async[RIO[R, _]]
     with ZioBlockingEnv[R, Throwable] {
 
--- a/zio-interop-cats/jvm/src/main/scala/zio/interop/ZioAsync.scala
+++ b/zio-interop-cats/jvm/src/main/scala/zio/interop/ZioAsync.scala
@@ -7,7 +7,7 @@ import zio.{ Promise, RIO, ZIO }
 import scala.concurrent.{ ExecutionContext, Future }
 
 private abstract class ZioAsync[R]
-    extends ZioTemporal[R, Throwable]
+    extends ZioTemporal[R, Throwable](Some(_))
     with Async[RIO[R, _]]
     with ZioBlockingEnv[R, Throwable] {
 
--- a/zio-interop-cats/shared/src/main/scala/zio/interop/cats.scala
+++ b/zio-interop-cats/shared/src/main/scala/zio/interop/cats.scala
@@ -91,17 +91,23 @@ abstract class CatsEffectInstances extends CatsZioInstances {
   implicit final def asyncRuntimeInstance[R](implicit runtime: Runtime[Clock & CBlocking]): Async[RIO[R, _]] =
     new ZioRuntimeAsync(runtime.environment)
 
-  implicit final def temporalRuntimeInstance[R, E](implicit runtime: Runtime[Clock]): GenTemporal[ZIO[R, E, _], E] =
-    new ZioRuntimeTemporal(runtime.environment)
+  implicit final def temporalRuntimeInstance[R, E](implicit
+    runtime: Runtime[Clock],
+    maybeEIsThrowable: Throwable =:= E = null
+  ): GenTemporal[ZIO[R, E, _], E] =
+    new ZioRuntimeTemporal(
+      runtime.environment,
+      if (maybeEIsThrowable != null) (e => Some(maybeEIsThrowable(e))) else (_ => None)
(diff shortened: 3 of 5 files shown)


joroKr21:
I'm not sure I like the fact that we sometimes promote the `Cause.Die` to an error (when the error type is `Throwable`). I've argued against it before, because in a generic context (where we use for example a generic `E` error type) we might not know that. But later instantiate `E =:= Throwable` anyway. So now we get different program behaviour depending on how we structured the code and I find that really surprising.

What is the motivation for that change? Any concrete use case? I guess it's not required for the cancellation fix.

neko-kai:
@joroKr21 
>I've argued against it before, because in a generic context (where we use for example a generic E error type) we might not know that. But later instantiate E =:= Throwable anyway. So now we get different program behaviour depending on how we structured the code and I find that really surprising.

One way around that is to just not have a generic E instances starting from `MonadCancel` – only `Throwable`. I did [just that](https://github.com/7mind/izumi/blob/develop/fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/Exit.scala#L206=) for BIO hierarchy's cats compat layer, because all the alternatives seem worse.

> What is the motivation for that change? Any concrete use case? I guess it's not required for the cancellation fix.

A lot of usages of `guaranteeCase`, even as simple as the following:

```scala
abc.guaranteeCase {
  case Outcome.Succeeded(_) =>
    log.info(s"Did abc")
  case errorOrCanceled => 
    log.crit(s"Couldn't finish abc because of $errorOrCanceled!")
}
```

End up broken when stuffing a die error into a `Succeeded` case. You can't match on the effect inside `Succeeded`, so you can't know if the effect actually finished or died without running the effect – in which case you won't be able to deal with the Die case anyway, because `handleErrorWith` won't catch it, and `guaranteeCase` will just return the Succeded outcome again and loop back.

Basically, whenever `guaranteeCase`/`bracketCase` is used, `Outcome.Succeeded` is used to tell whether the effect actually ran to completion or not – this is critical in finalizers, otherwise these functions won't be required at all. `Outcome` doc uses OptionT as a motivating example for using an effect value in Succeeded – but it also illustrates the problem – monad transformers obliterate `bracketCase`/`guaranteeCase` usages with their additional error channel, because they make the error in the inner monad masquerade as a success in the base monad. Lack of this problem is a huge advantage of using a bifunctor effect.

`Outcome.Canceled` is also not a good candidate to convert a Die to.

Another reason is consistency with CE2 version of interop, where `Die` is `ExitCase.Error`. There was no choice back then because ExitCase forced a Throwable error anyway. But, I don't think the additional type parameter on Outcome changed the situation much – we're either forced to pretend that ZIO is EitherT and force the same usability problems downstream by stuffing untyped errors into Succeeded, or pin the error type to Throwable and get back the CE2 ExitCase anyway.

* One other compromise solution to allow generic a error type in `GenTemporal` and `GenConcurrent` is to make the operated error a `Cause[E]` -
  ```scala
  implicit final def concurrentInstance[R, E]: GenConcurrent[ZIO[R, E, _], Cause[E]] = ...
  ```
  That is, in this interpretation `handleErrorWith` = `ZO.catchAllCause`. This both allows to catch defects in CE and actually makes use of the `E` parameter in Outcome to represent all the possible errors in ZIO instead of masking them as successes.

joroKr21:
@neko-kai you summarised exactly what I was thinking - this is no different than `EitherT` which to me is a tool to say - in this part of my program I care about a specific type of errors and the rest are invisible. In a similar way that you can have `MonadError` based on a prism in the meow-mtl library.

I also realised we are missing the `MonadError` instance for `Cause` and I wonder if we can provide it in a way that avoids conflicts with the rest 🤔

But ok in general if you see the users asking for this maybe we should make this compromise - I don't know if there is a good way to document this to avoid surprises.

calvinlfer (approved):

adamgfraser (approved):

neko-kai:
@joroKr21 @calvinlfer @adamgfraser 

I changed the PR to split off non-Throwable instances into a separate object, with these instances' MonadError part now now on Cause[E] error instead of E error. This allows consistent conversion of all non-interrupt errors into `Outcome.Errored` for both generic and Throwable instances, but this also means that generic instances are more powerful than default instances – they can catch defects etc.:

```
object generic {
  implicit final def concurrentInstanceCause[R, E]: GenConcurrent[ZIO[R, E, _], Cause[E]] = ...
  implicit final def temporalInstanceCause[R, E]: GenConcurrent[ZIO[R, E, _], Cause[E]] = ..
}
```

joroKr21:
> I changed the PR to split off non-Throwable instances into a separate object, with these instances' MonadError part now operating on Cause[E] instead E errors. This allows consistent conversion of all non-interrupt errors into Outcome.Errored for both generic and Throwable instances, but this also means that generic instances are more powerful than default Throwable instances – they can catch defects etc.:

Wait, does that mean we lose the generic `MonadError` instances for `ZIO[_, E, A]`?
I fear that could break too much existing code.

PR #544, opened 2022-06-16
Fix #503 implement `MonadCancel#canceled` by sending an external interrupt to current fiber via `Fiber.unsafeCurrentFiber`

### LINKED ISSUE #503: MonadCancel instance is unlawful
The behavior of `MonadCancel#canceled` method doesn't seem to map to `ZIO.interrupt` at all. In particular the following law:
```scala
  def uncancelableCanceledAssociatesRightOverFlatMapAttempt[A](fa: F[A]) =
    (F.uncancelable(_ => F.canceled >> fa).attempt >> F.unit) <->
      F.forceR(F.uncancelable(_ => fa))(F.canceled)
```
and the associated comment:
```
   * In the following example, the fiber requests self-cancelation in a masked
   * region, so cancelation is suppressed until the fiber is completely
   * unmasked. `fa` will run but `fb` will not.
   *
   * {{{
   *
   *   F.uncancelable { _ =>
   *     F.canceled *> fa
   *   } *> fb
   *
   * }}}
```

This appears to mean that:

1. `canceled` is ignored in uninterruptible regions
2. However, when uncancelable region is over, `canceled` will execute immediately (hence 'associates right')

That behaves just like external interruption - but sent from yourself to yourself, not the immediate internal interruption of `ZIO.interrupt`.
Implementing that directly would require calling `interruptAs` on the `FiberContext` of the current fiber, which I'm not sure is doable without modifying the runtime.

Note: tests did not catch that since Eq instance for IO in tests only compares output Cause, not executed effects – which would be `Cause.Interrupted` on both sides of the equation, but the `fa` effect will only run on the right hand side.

/cc @adamgfraser @joroKr21

neko-kai:
* Fix #503 implement `MonadCancel#canceled` by sending an external interrupt to current fiber via `Fiber.unsafeCurrentFiber`
* Workaround zio/zio#6911 Only consider the *current* fiber being interrupted as `Outcome.Canceled`
* Consider internal interruption & Die to be `Outcome.Errored`
* Trigger `MonadCancel#onCancel` only if the *current* fiber is being interrupted
* Only trigger `MonadCancel#onCancel` on interruption, instead of any Exit status that's not a typed error
* Implement `MonadCancel#bracketFull`, `bracketCase` & `guaranteeCase`
* Redefine `toEffect` using `unsafeRunAsyncCancelable` to avoid getting a `BoxedException` leaking from underlying Future when `MonadCancel#canceled` is used inside the converted effect.
* Change the implementation of Async#async to follow `async left is uncancelable sequenced raiseError` law.

  Consider the law code:

  ```scala
  // format: off
  def asyncLeftIsUncancelableSequencedRaiseError[A](e: Throwable, fu: F[Unit]) =
    (F.async[A](k => F.delay(k(Left(e))) >> fu.as(None)) <* F.unit) <-> (F.uncancelable(_ => fu) >> F.raiseError(e))
  // format: on
  ```

  if `fu` is `F.never`, then an implementation based on ZIO.effectAsyncM can never satisfy this law,
because it runs the register code on a separate fiber and ignores its effect.
The law clearly states that register effect must run first and must run on the same fiber
as the callback listener, so as to supercede it.

### DIFF
--- a/zio-interop-cats-tests/jvm/src/test/scala/zio/interop/CatsInteropSpec.scala
+++ b/zio-interop-cats-tests/jvm/src/test/scala/zio/interop/CatsInteropSpec.scala
@@ -86,15 +86,17 @@ object CatsInteropSpec extends CatsRunnableSpec {
       for {
         counter <- F.ref("")
         _       <- F.guaranteeCase(
-                     F.onCancel(
-                       ZIO.collectAllPar(
-                         List(
-                           ZIO.unit.forever,
-                           counter.update(_ + "A") *> ZIO.die(new RuntimeException("x")).unit
-                         )
-                       ),
-                       counter.update(_ + "1")
-                     )
+                     F.onError(
+                       F.onCancel(
+                         ZIO.collectAllPar(
+                           List(
+                             ZIO.unit.forever,
+                             counter.update(_ + "A") *> ZIO.die(new RuntimeException("x")).unit
+                           )
+                         ),
+                         counter.update(_ + "1")
+                       )
+                     ) { case _ => counter.update(_ + "B") }
                    ) {
                      case Outcome.Errored(_)   => counter.update(_ + "C")
                      case Outcome.Canceled()   => counter.update(_ + "2")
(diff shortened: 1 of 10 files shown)


neko-kai:
@joroKr21 @adamgfraser 
This is done, but includes commits from https://github.com/zio/interop-cats/pull/543 until the latter is merged.
Unfortunately, enabling generation of non-Succeed IOs still surfaces more law failures that I haven't tackled in this PR https://github.com/zio/interop-cats/pull/544/commits/8c8438421d1e1e4f3e3e6a6aee2dcba18d55f370

adamgfraser (approved):

PR #545, opened 2022-06-24
ZIO 2.0.0

vigoo:


zio-interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -177,21 +180,23 @@ private class CatsConcurrentEffect[R](rts: Runtime[R])
...
-              }
-          )(_ => fa)
-          .interruptible
-          .forkDaemon
-          .map(_.interrupt.unit)
+            )(_ => fa)
+            .interruptible
+            .forkDaemon
+            .map(_.interrupt.unit)
+        }.getOrThrowFiberFailure()
```
codingismy11to7, line 199: not an expert, but i think maybe this should be `.getOrThrow()` since we're limited to E of Throwable?

zio-interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -165,8 +165,11 @@ private class CatsConcurrentEffect[R](rts: Runtime[R])
     implicit def trace: Trace = InteropTracer.newTrace(cb)
 
     effect.SyncIO {
-      rts.unsafeRunAsyncWith(fa.exit) { exit =>
-        cb(exit.flatMap(identity).toEither).unsafeRunAsync(_ => ())
+      Unsafe.unsafeCompat { implicit u =>
+        val fiber = rts.unsafe.fork(fa.exit)
+        fiber.unsafe.addObserver { exit =>
+          cb(exit.getOrThrowFiberFailure().toEither).unsafeRunAsync(_ => ())
+        }
```
codingismy11to7, line 172: i get the same test results with the (slightly simpler)
```scala
      Unsafe.unsafe { implicit u =>
        rts.unsafe.fork(fa).unsafe.addObserver { exit =>
          cb(exit.toEither).unsafeRunAsync(_ => ())
        }
```
but again, not an expert

zio-interop-cats/shared/src/main/scala/zio/interop/cats.scala
```diff
@@ -237,8 +242,8 @@ private class CatsConcurrent[R] extends CatsMonadError[R, Throwable] with Concur
       fc.interruptible
 
     (run(fa) raceWith run(fb))(
-      { case (l, f) => l.fold(f.interrupt *> ZIO.failCause(_), ZIO.succeedNow).map(lv => Left((lv, toFiber(f)))) },
-      { case (r, f) => r.fold(f.interrupt *> ZIO.failCause(_), ZIO.succeedNow).map(rv => Right((toFiber(f), rv))) }
+      { case (l, f) => l.foldExit(f.interrupt *> ZIO.failCause(_), ZIO.succeedNow).map(lv => Left((lv, toFiber(f)))) },
+      { case (r, f) => r.foldExit(f.interrupt *> ZIO.failCause(_), ZIO.succeedNow).map(rv => Right((toFiber(f), rv))) }
```
codingismy11to7, line 246: curious here as to what the difference is between `foldExit` and `foldCauseZIO` since they both compile and run with the same test results (looks like foldExit might save an instantiation?)

adamgfraser: `foldExit` preserves the type information that the value is an `Exit` and not some other subtype of `ZIO`.

codingismy11to7:
i had made mostly the same changes (except explicitly marking the Schedule methods that broke compilation with @nowarns) and just checked out this branch and noticed that 6 `catzSpec` laws tests are failing the same as mine; just checked and verified that they pass with the released 2-RC6 version. they seem to be issues with interruptibility that are above my pay grade

@adamgfraser since you're active ATM that's another question - got warnings about schedule combinators making the compiler infer types as Any, seems to maybe be a 2.0.0 rough spot

adamgfraser:
@codingismy11to7 I would recommend turning off that warning. It can trigger spurious warnings with contravariant types like the ZIO environment as well. We have disabled it in most ZIO projects.

zio-interop-cats/shared/src/main/scala/zio/interop/cats.scala
codingismy11to7: ah okay, i missed that Exit is a ZIO and that foldExit can only be called on an Exit...but since it's a fold with inputs being a Cause or an A I don't see how it preserves type information. but probably something i'm missing, not asking for an explanation at midnight.

adamgfraser: `foldCauseZIO` returns a `ZIO` versus `foldExit` returns an `Exit`. So if use `foldCauseZIO` we "forget" that this is a workflow that is "done" and we can just get the value from.

adamgfraser: Also notice how `foldExit` takes a function that returns a `Z` and returns a `Z` and `foldCauseZIO` takes a function that returns a `ZIO` and returns a `ZIO`. So they are going to be the same when you are returning a `ZIO` but `foldExit` allows you to actually get a value out of a `ZIO` in the case where it is an `Exit`.

zio-interop-cats/shared/src/main/scala/zio/interop/cats.scala
vigoo: It was `unsafeRun` before, which was defined as:

```scala
  final def unsafeRun[E, A](zio: ZIO[R, E, A])(implicit trace: Trace): A =
    unsafeRunSync(zio).getOrElse(c => throw FiberFailure(c))
```

so this way we keep the same behavior as before

adamgfraser (approved):

Issue #562, opened 2022-07-12
`MonadCancel#uncancelable` disrepancies


neko-kai:
Given the description of behavior of `Poll` produced by `MonadCancel#uncancelable` in its scaladoc:

```scala
 /** ...
   * Masks can also be stacked or nested within each other. If multiple masks are active, all
   * masks must be undone so that cancelation can be observed. In order to completely unmask
   * within a multi-masked region the poll corresponding to each mask must be applied to the
   * effect, outermost-first.
   *
   * {{{
   *
   *   F.uncancelable { p1 =>
   *     F.uncancelable { p2 =>
   *       fa *> p2(p1(fb)) *> fc
   *     }
   *   }
   *
   * }}}
   *
   * The following operations are no-ops:
   *
   *   1. Polling in the wrong order
   *   1. Subsequent polls when applying the same poll more than once: `poll(poll(fa))` is
   *      equivalent to `poll(fa)`
   *   1. Applying a poll bound to one fiber within another fiber
   *
   * @param body
   *   A function which takes a [[Poll]] and returns the effect that we wish to make
   *   uncancelable.
```

`ZIO.uninterruptibleMask` as `uncancelable` does not satisfy this description:

1. In the code example, with ZIO, applying `p1(fb)` alone is enough to restore full interruptibility. The sentence "In order to completely unmask within a multi-masked region the poll corresponding to each mask must be applied to the effect, outermost-first." is not satisfied.
2. "Applying a poll bound to one fiber within another fiber is no-op" is not satisfied – restore functions produced by ZIO.uninterruptibleMask can migrate fibers and are not invalidated when used on another fiber.

PR #595, opened 2022-08-05
Fix unlawful instances

johnspade:
Port to ZIO 2 and apply the following fixes by @neko-kai from the `master` branch:

- #542 
- #543 
- #544 
- #549

### DIFF
--- a/zio-interop-cats-tests/shared/src/test/scala/zio/interop/CatsSpecBase.scala
+++ b/zio-interop-cats-tests/shared/src/test/scala/zio/interop/CatsSpecBase.scala
@@ -94,7 +94,12 @@ private[zio] trait CatsSpecBase
         FiberRef.currentBlockingExecutor -> ::(fiberId -> blockingExecutor, Nil)
       )
     )
-    val runtimeFlags     = RuntimeFlags.default
+    val runtimeFlags     = RuntimeFlags( // todo
+      RuntimeFlag.FiberRoots,
+      RuntimeFlag.Interruption,
+      RuntimeFlag.CooperativeYielding,
+      RuntimeFlag.CurrentFiber
+    )
     Runtime(ZEnvironment.empty, fiberRefs, runtimeFlags)
   }
 
@@ -113,6 +118,8 @@ private[zio] trait CatsSpecBase
   implicit val eqForNothing: Eq[Nothing] =
     Eq.allEqual
 
+  // workaround for laws `evalOn local pure` & `executionContext commutativity`
+  // (ZIO cannot implement them at all due to `.executor.asEC` losing the original executionContext)
   implicit val eqForExecutionContext: Eq[ExecutionContext] =
     Eq.allEqual
 
--- a/zio-interop-cats-tests/shared/src/test/scala/zio/interop/CatsSpecBase.scala
+++ b/zio-interop-cats-tests/shared/src/test/scala/zio/interop/CatsSpecBase.scala
@@ -4,7 +4,7 @@ import cats.effect.testkit.TestInstances
 import cats.effect.kernel.Outcome
 import cats.effect.IO as CIO
 import cats.syntax.all.*
-import cats.{ Eq, Order }
+import cats.{ Eq, Id, Order }
 import org.scalacheck.{ Arbitrary, Cogen, Gen, Prop }
 import org.scalatest.funsuite.AnyFunSuite
 import org.scalatest.prop.Configuration
@@ -69,15 +69,18 @@ private[zio] trait CatsSpecBase
       ???
   }
 
-  def unsafeRun[E, A](io: IO[E, A])(implicit ticker: Ticker): Exit[E, Option[A]] =
+  def unsafeRun[E, A](io: IO[E, A])(implicit ticker: Ticker): (Exit[E, Option[A]], Boolean) =
     try {
       var exit: Exit[E, Option[A]] = Exit.succeed(Option.empty[A])
+      var interrupted: Boolean     = true
       Unsafe.unsafeCompat { implicit u =>
-        val fiber = runtime.unsafe.fork[E, Option[A]](io.asSome)
+        val fiber = runtime.unsafe.fork[E, Option[A]](signalOnNoExternalInterrupt(io)(ZIO.succeed {
+          interrupted = false
+        }).asSome)
         fiber.unsafe.addObserver(exit = _)
       }
       ticker.ctx.tickAll(FiniteDuration(1, TimeUnit.SECONDS))
-      exit
+      (exit, interrupted)
     } catch {
       case error: Throwable =>
         error.printStackTrace()
@@ -124,22 +127,15 @@ private[zio] trait CatsSpecBase
     Eq.allEqual
 
   implicit val eqForCauseOfNothing: Eq[Cause[Nothing]] =
-    eqForCauseOf[Nothing]
-
-  implicit def eqForCauseOf[E]: Eq[Cause[E]] =
-    (x, y) => (x.isInterrupted && y.isInterrupted) || x == y
-
-  implicit def eqForExitOfNothing[A: Eq]: Eq[Exit[Nothing, A]] = {
-    case (Exit.Success(x), Exit.Success(y)) => x eqv y
-    case (Exit.Failure(x), Exit.Failure(y)) => x eqv y
-    case _                                  => false
-  }
+    (x, y) => (x.isInterrupted && y.isInterrupted && x.failureOption.isEmpty && y.failureOption.isEmpty) || x == y
 
   implicit def eqForUIO[A: Eq](implicit ticker: Ticker): Eq[UIO[A]] = { (uio1, uio2) =>
-    val exit1 = unsafeRun(uio1)
-    val exit2 = unsafeRun(uio2)
-    (exit1 eqv exit2) || {
-      println(s"$exit1 was not equal to $exit2")
+    val (exit1, i1) = unsafeRun(uio1)
+    val (exit2, i2) = unsafeRun(uio2)
+    val out1        = toOutcomeCauseOtherFiber[Id, Nothing, Option[A]](i1)(identity, exit1)
+    val out2        = toOutcomeCauseOtherFiber[Id, Nothing, Option[A]](i2)(identity, exit2)
+    (out1 eqv out2) || {
+      println(s"$out1 was not equal to $out2")
       false
     }
   }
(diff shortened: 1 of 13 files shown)


johnspade:
@neko-kai please help with the failing tests 🙏

neko-kai:
@johnspade 
Hi, I fixed one of the failing tests, CatsInteropSpec, but there are still failures that surface when `genOfRace`/`genOfParallel` or `genCancel` generators are enabled. I'm not sure if these are caused by more uncovered behavior or bugs in ZIO 2.
I would suggest to investigate errors `genCancel` with genRace & genParallel disabled

johnspade:
@neko-kai 
These tests are failing with enabled `genCancel` and disabled `genRace` & `genParallel`:
- Temporal[Task].temporal.onCancel associates over uncancelable boundary
- GenSpawn[IO[Int, _], Cause[Int]].concurrent.onCancel associates over uncancelable boundary
- MonadCancel[IO[In t, _], Cause[Int]].monadCancel.onCancel associates over uncancelable boundary
- Async[Task].async.onCancel associates over uncancelable boundary

It looks like we have to fix #562.

neko-kai:
@johnspade 
Yes, I've seen them fail with genCancel & disabled genRace/genParallel. There are no laws that check for issues in #562, so I doubt that's it. By investigating I mean using debugger & prints (and possibly overriding law code to insert debugging or rewriting the law in pure ZIO to inspect it closely) - to obtain understanding of what actually fails exactly. I suspect ZIO 2 has bugs which cause these failures, because as far as I see you copied the fixes in ZIO1 correctly.

johnspade:
The failing case for the "monadCancel.onCancel associates over uncancelable boundary" law looks like this in pure ZIO:
```scala
def canceled = {
  def loopUntilInterrupted: UIO[Unit] =
    ZIO.descriptorWith(d => if (d.interrupters.isEmpty) ZIO.yieldNow *> loopUntilInterrupted else ZIO.unit)

  for {
    _ <- ZIO.withFiberRuntime[Any, Nothing, Unit]((thisFiber, _) => thisFiber.interruptAsFork(thisFiber.id))
    _ <- loopUntilInterrupted
  } yield ()
}

ZIO.unit.onExit(_ => canceled *> ZIO.never)
```

This code hangs and fails the test, but works fine with `ZIO.interrupt`. It means that this Cats code won't work correctly:
```scala
F.onCancel(F.uncancelable(_ => fa), F.canceled *> F.never)
```

`thisFiber.interruptAsFork` can't interrupt itself inside `onExit` because it's an uninterruptible region. @neko-kai any ideas?

neko-kai:
@johnspade 
You probably found the problem. This code shouldn't hang on ZIO1. Check this line https://github.com/zio/interop-cats/pull/595/files#diff-e27c0f2c26e6b28ceb1612ce7fe2ea034311be23ed842e07678687c6b4c6459fR280 - the `else ZIO.unit` clause is not actually dead code - with ZIO1 `interrupters` will be filled even when we're in an uninterruptible region, so `canceled` will return Unit instead of hanging when uninterruptible.

If this hangs on ZIO2 that means it works differently now - we may need to find another way to reliably check for external interrupters (in uninterruptible region). /cc @adamgfraser 

Or we could just give up and return Unit immediately if we're in an uninterruptible region.

adamgfraser:
Don't we just need to use `ZIO.interrupt`?

```scala
object Example extends ZIOAppDefault {

  val awaitInterruption: UIO[Any] =
    ZIO.descriptorWith(d => if (d.interrupters.nonEmpty) ZIO.interrupt *> ZIO.never else awaitInterruption)

  val run =
    for {
      promise <- Promise.make[Nothing, Unit]
      fiber   <- ZIO.unit.onExit(_ => promise.succeed(()) *> awaitInterruption).forkDaemon
      _       <- promise.await
      exit    <- fiber.interrupt
      _       <- Console.printLine(exit)
    } yield ()
}
```

neko-kai:
Cats `canceled` shouldn't interrupt immediately, but return Unit and delay interruption until the uninterruptible region is over, so that's correct that we don't use `ZIO.interrupt`

adamgfraser:
Okay, then I think we want something like:

```scala
object Example extends ZIOAppDefault {

  val canceled: UIO[Any] =
    ZIO.fiberIdWith { fiberId =>
      FiberRef.interruptedCause.update(_ ++ Cause.interrupt(fiberId))
    }

  val run =
    for {
      fiber <- ZIO.unit.onExit(_ => canceled *> ZIO.debug("Still going")).forkDaemon
      exit  <- fiber.await
      _     <- Console.printLine(exit)
    } yield ()
}
```

johnspade:
>Actually, looking at the examples again, they should hang anyway because of *> never.

Yeah, sorry, I got confused, this is working as intended.

jdegoes (approved):

zio-interop-cats-tests/shared/src/test/scala/zio/interop/CatsSpecBase.scala
```diff
@@ -113,44 +117,62 @@ private[zio] trait CatsSpecBase
...
+    val (exit1, i1) = unsafeRun(uio1)
+    val (exit2, i2) = unsafeRun(uio2)
+    val out1        = toOutcomeCauseOtherFiber[Id, Nothing, Option[A]](i1)(identity, exit1)
+    val out2        = toOutcomeCauseOtherFiber[Id, Nothing, Option[A]](i2)(identity, exit2)
+    (out1, out2) match {
+      case (Succeeded(Some(a)), Succeeded(Some(b)))          => a eqv b
+      case (Succeeded(Some(_)), _) | (_, Succeeded(Some(_))) =>
+        println(s"$out1 was not equal to $out2")
+        false
+      case _                                                 => true
```
johnspade, line 138: We can ignore results like `Succeeded(None)`, it means that IO under tests did not produce a result (`IO.never`, for example)

johnspade:
@neko-kai Please take a look if everything is okay in this PR, I can fix it in the separate PR

johnspade:
@adamgfraser I've tried your implementation for `canceled` but unfortunately some tests are failing (I didn't check why)

adamgfraser:
@johnspade Okay. Is this still an issue? Can we isolate what the expected behavior is that we need to support?

johnspade:
@adamgfraser No, everything is working fine now (at least according to law tests). I was wrong and my example should hang with Cats.

> Cats canceled shouldn't interrupt immediately when in uninterruptible region, but return Unit and delay interruption until the uninterruptible region is over

This is the expected behavior.

adamgfraser:
@johnspade Okay, great!

zio-interop-cats-tests/shared/src/test/scala/zio/interop/CatsSpecBase.scala
neko-kai: @johnspade We can't ignore it. If both lhs and rhs hang then it's fine, but not if one of the sides finishes and the other hangs. And `case _ => true` here ignores everything, not just `Succeeded(None)`

neko-kai:
@adamgfraser @jdegoes @johnspade 
The expected behavior is for both sides of the equation to hang, not just one of them. This line - https://github.com/zio/interop-cats/pull/595/commits/f13c6c03d93ea523ae9b0c20750ae110c237e9a6#diff-b8ed2c048152009ccfdae3f8027be0a596b699ca0785aa61dc72000427289372R138 - doctors the test results and makes nearly everything pass. The laws are still not fixed. I would revert the merge in this case tbh

johnspade:
```scala
F.uncancelable(_ => F.onCancel(fa, fin)) <-> F.onCancel(F.uncancelable(_ => fa), fin)
```
When the `genCancel` generator is used for the left side and the `genNever` generator is used for the right side, they can't be equal. The left side will not hang in this case.

neko-kai:
@johnspade 
The same generator is used on both sides, since `fa` is the result of generator

johnspade:
@neko-kai I've observed such behavior in my debugging, I guess generators are randomly chosen on each `fa` evaluation

zio-interop-cats-tests/shared/src/test/scala/zio/interop/CatsSpecBase.scala
johnspade: Well, cats-effect uses similar relaxed `Eq` for law testing

neko-kai:
@johnspade In that case no tests would ever pass at all, because e.g. `genSuccess` and `genFail` would never be equal and also there would be different values inside `genSuccess`. This is also structurally impossible because we only have `Gen[IO[A]]` not `IO[Gen[A]]` - the output IO is determined once the generator is run.

Also, the tests also fail when using `catsConversionGenerator` in `GenIOInteropCats` - since that converts from cats IO generator - that works with cats IO laws, it shouldn't have the non-determinism you describe.

Also it works on ZIO1 branch which has about the same generators and eq instances.

What you could have observed is `genParallel`/`genRace` running effects in indeterminate order, however the `fa` is guaranteed to be the same exact instance and not to have random generators inside it that would make it different when run.

zio-interop-cats-tests/shared/src/test/scala/zio/interop/CatsSpecBase.scala
neko-kai: @johnspade
I would *really* love to know how you came up with that, but cats-effect tests use the `eqIOA` instance defined in `TestInstances` - https://github.com/typelevel/cats-effect/blob/series/3.x/tests/shared/src/test/scala/cats/effect/IOSpec.scala#L1580 - which uses the `Eq[Outcome[Option, Throwable, A]]` instance - i.e. `out1 eqv out2` here - which checks all the branches not just success.

```scala
  implicit def eq[F[_], E: Eq, A](implicit FA: Eq[F[A]]): Eq[Outcome[F, E, A]] =
    Eq instance {
      case (Canceled(), Canceled()) => true
      case (Errored(left), Errored(right)) => left === right
      case (Succeeded(left), Succeeded(right)) => left === right
      case _ => false
    }
```

adamgfraser:
@neko-kai I'm fine with reverting this until we are all comfortable with the changes here since it seems like we are still figuring this out.

@johnspade I will take a look myself but did my suggestion fix the issue with this test? I know you said it caused issues but I think we need to understand what those issues are.

johnspade:
Okay, let's revert and reopen it

johnspade:
@adamgfraser No, this test (monadCancel.onCancel associates over uncancelable boundary) is still failing with your suggestions

adamgfraser:
@johnspade Okay let's take a look at that.

PR #602, opened 2022-08-18
Bring back interoperability between ZManaged and Resource for Cats-Effect 2.

Wosin:
This is pretty much just bringing back what was already defined for CE 2. I didn't wanted to split that into multiple classes to make it the same as it's currently defined for CE3.


adamgfraser:
@Wosin Going to close this for now but feel free to open again if you would like to resolve the test failures and get this in.

PR #646, opened 2023-01-16
Upgrade ZIO Version

adamgfraser:


calvinlfer (approved):

neko-kai:
Was `ZIO#raceWith` deleted?

adamgfraser:
`ZIO.raceWith` is implemented in terms of `ZIO.raceFibersWith` which forks the child fibers in the parent scope and is inconsistent with Cats Effect semantic. So we implement our own version of `raceFibersWith` that does not do this and implement our own version of `raceWith` in terms of that.

vigoo:
Maybe it would worth marking the difference with a comment, I was looking at the two for minutes and did not see it :D

adamgfraser:
Added a comment.

neko-kai (approved):

PR #654, opened 2023-02-12
Removes Usages Of ZIO.succeedNow

adamgfraser:
This is an internal API that will be deleted at some point.


neko-kai:
@adamgfraser Does that apply to `zio.internal.ZIOSucceedNow` too or we could just call that instead?

adamgfraser:
@neko-kai I don't think that exists on ZIO 2.

neko-kai (approved):

PR #661, opened 2023-04-22
Hide warning in Concurrent.racePair

danilbykov:
Such code
```scala
object Test extends zio.ZIOAppDefault {

  def failure[F[_]: Concurrent]: F[Int] =
    Concurrent[F].raiseError[Int](new IOException("!"))

  def race[F[_]: Concurrent: Timer]: F[Option[Int]] =
    Concurrent.timeoutTo[F, Int](failure, 3.second, Concurrent[F].raiseError(new NullPointerException()))
      .map(i => Option(i))
      .recover {
        case _: IOException => None
      }

  override val run =
    race[zio.Task].debug("result")
}
```

generates following message in log with DEBUG level
```json
{
  "@timestamp":"2023-04-22T16:19:34",
  "level":"DEBUG",
  "thread":"zio-fiber-11",
  "location":"zio.interop.CatsConcurrent.racePair.trace:240",
  "file":"cats.scala",
  "cause":"Exception in thread \"zio-fiber-11\" java.io.IOException: !
	at test.Test$.failure(Test.scala:29)
	at test.Test$.race(Test.scala:35)
	at test.Test$.<clinit>(Test.scala:51)
	at test.Test.main(Test.scala)
	at zio.interop.CatsMonadError.raiseError(cats.scala:336)
	at zio.interop.CatsConcurrent.racePair.trace(cats.scala:240)",
  "message":"Fiber zio-fiber-11 did not handle an error"
}
```

Similar code without tagless final doesn't print any warnings. It looks like it is misleading message and it should be hidden in Concurrent.racePair.

### DIFF
--- a/zio-interop-cats/shared/src/main/scala/zio/interop/cats.scala
+++ b/zio-interop-cats/shared/src/main/scala/zio/interop/cats.scala
@@ -239,12 +239,21 @@ private class CatsConcurrent[R] extends CatsMonadError[R, Throwable] with Concur
   ): RIO[R, Either[(A, effect.Fiber[RIO[R, *], B]), (effect.Fiber[RIO[R, *], A], B)]] = {
     implicit def trace: Trace = CoreTracer.newTrace
 
-    def run[C](fc: RIO[R, C]): ZIO[R, Throwable, C] =
-      fc.interruptible
+    def run[C](fc: RIO[R, C]): URIO[R, Exit[Throwable, C]] =
+      fc.exit.interruptible
 
     (run(fa) raceWith run(fb))(
-      { case (l, f) => l.foldExit(f.interrupt *> ZIO.failCause(_), ZIO.succeed(_)).map(lv => Left((lv, toFiber(f)))) },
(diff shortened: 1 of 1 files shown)


adamgfraser (changes requested):
We should not be changing the behavior of `racePair` here. The child fiber failed and its failure should be reported at the debug level. This is the same behavior as ZIO. For example:

```scala
object Example extends ZIOAppDefault {

  val debugLogger =
    ZLogger.default.map(println(_))

  override val bootstrap =
    Runtime.addLogger(debugLogger)

  val failure =
    ZIO.fail(new IOException("!"))

  val race =
    failure.race(ZIO.sleep(3.seconds)).ignore

  val run =
    race.debug
}
```

Particular operators such as timing out that want to handle the possibility of the effect failing before forking it are free to do so.

danilbykov:
Hmm, yes,
```scala
failure.race(ZIO.sleep(3.seconds)).ignore
```
prints warning but this code doesn't
```scala
failure.timeoutFail(new NullPointerException())(3.seconds).ignore
```

So it looks like it is implementation detail which in general case means nothing and may be misleading.

Issue #687, opened 2024-02-01
Memory leaks while using fs2-kafka consumer


vkorchik:
I have a ZIO-based app but for kafka I am still using fs2-kafka. As of interops I was using `zio-interop-cats:23.0.0.0`. Everything was fine.

After bumping all the libs to the newest versions I got memory leak. After some investigation I thought memory leak could be due to inconsistency of `cats-effect` versions: the newest one is 3.5.3 but `zio-interops-cats` are compiled against 3.4.8. Starting from 3.5.3 there are [some crucial changes](https://github.com/typelevel/cats-effect/releases/tag/v3.5.0) in runtime semantics so I have rollbacked versions a bit so they are the same throughout the whole stack, but memory leak still appears.

I have examined different versions of cats/interops/zio but no success. 

After even more debugging I see that **the consumer side is leaking**: once consumer consumes events it does something wrong. A lot of objects of `zio.ZIO$EvaluationStep$UpdateRuntimeFlags$$anon$28` are created, but never released. Number is just growing.

Versions that are working fine:
- zio:2.0.1
- zio-interop-cats:23.0.0.0
- cats-effect:3.3.13
- fs2-kafka:2.5.0

Versions that I am testing against and getting leaks:
- zio:2.0.16
- zio-interop-cats:23.1.0.0
- cats-effect:3.4.8
- fs2-kafka:3.0.0

Versions that are leaking are many but basically all of them are based on `cats-effect:3.4.x`.

This is the link to [the repo](https://github.com/vkorchik/leaking-zio-interop-fs2-kafka) and to [the heap dump](https://drive.google.com/file/d/13WTuRHbuRmOUOTus2b3_vtXxYPzYmS4Q/view?usp=sharing) from leaking run.

And also picture of heap usage. Once heap is close to the current max size of heap it just breaks through it and this never ends until all heap is used.
![image](https://github.com/zio/interop-cats/assets/31250785/a34aa89e-a38b-4ea7-9045-e98be5dea584)

Appreciate any help regarding this issue.

vkorchik:
After some more tests I have found out that problem started at `zio-interop-cats:23.0.04` (and above). 23.0.0.3 works fine for no matter which version of `zio` as well as `fs2-kafka`.

vkorchik:
After even more tests I have found the commit/PR and the exact line of code that has broken cats-interop back then and my example as well. It was one of the PRs of 2.0.11 release.
[...]

PR #688, opened 2024-02-17
Fix memory leak issue in ZIOConcurrent

vkorchik:
Potential fix for https://github.com/zio/interop-cats/issues/687


vkorchik:
Memory usage looks like this:
image

adamgfraser (approved):

PR #691, opened 2024-04-17
Use `FiberRuntime#unsafe.addObserver` for bincompat with v2.1

kyri-petrou:
To avoid bincompat issues with 2.1.x versions  (see https://github.com/zio/zio/pull/8757)


kyri-petrou:
After discussing this with @ghostdogpr, might be better to fix the bincompat issues in zio altogether

PR #696, opened 2024-07-25
Fix propagation of interruption from CIO to ZIO

### LINKED ISSUE #695: Canceler effect in toEffect does not run
I could be misunderstanding but this doesn't look right:
https://github.com/zio/interop-cats/blob/879070e7f9d2310f21734f81b4d42d35030acc05/zio-interop-cats/shared/src/main/scala/zio/interop/package.scala#L46-L48

That returns a `UIO` but never runs it.

kyri-petrou:
/fixes #695

/claim #695

Note that I also changed the code to use `runOrFork` instead of `fork` within `toEffect`. This is to avoid the overhead of forking fibers when the ZIO effect can be evaluated synchronously. Users that convert simple ZIO effects to CIO often will see a very notable performance improvement from this change

### DIFF
--- a/zio-interop-cats/shared/src/main/scala/zio/interop/package.scala
+++ b/zio-interop-cats/shared/src/main/scala/zio/interop/package.scala
@@ -35,23 +35,23 @@ package object interop {
     F.defer {
       val interrupted = new AtomicBoolean(true)
       F.asyncCheckAttempt[Exit[Throwable, A]] { cb =>
-        Unsafe.unsafe { implicit unsafe =>
-          val fiber = R.unsafe.runOrFork {
+        F.delay {
+          implicit val unsafe: Unsafe = Unsafe.unsafe
+
+          val out = R.unsafe.runOrFork {
             signalOnNoExternalInterrupt {
               rio
             }(ZIO.succeed(interrupted.set(false)))
           }
-          val out   = fiber match {
-            case Left(fib) =>
-              fib.unsafe.addObserver(exit => cb(Right(exit)))
+          out match {
+            case Left(fiber) =>
+              fiber.unsafe.addObserver(exit => cb(Right(exit)))
               Left(Some(F.async_[Unit] { cb =>
-                fib.unsafe.addObserver(_ => cb(Right(())))
-                fib.tellInterrupt(Cause.interrupt(fib.id))
+                fiber.unsafe.addObserver(_ => cb(Right(())))
+                fiber.tellInterrupt(Cause.interrupt(fiber.id))
               }))
-            case Right(v)  => Right(v) // No need to invoke the callback, sync resumption will take place
+            case Right(v)    => Right(v) // No need to invoke the callback, sync resumption will take place
           }
-
-          F.pure(out)
         }
       }.flatMap { exit =>
         toOutcomeThrowableOtherFiber(interrupted.get())(F.pure(_: A), exit) match {
--- a/zio-interop-cats/shared/src/main/scala/zio/interop/package.scala
+++ b/zio-interop-cats/shared/src/main/scala/zio/interop/package.scala
@@ -54,7 +54,7 @@ package object interop {
                   fiber.unsafe.removeObserver(completeCb)
                   fiber.tellInterrupt(Cause.interrupt(fiber.id))
                   // Allow the interruption to be interrupted
-                  Some(fiber.unsafe.removeObserver(interruptCb))
+                  Some(F.delay(fiber.unsafe.removeObserver(interruptCb)))
                 }
               }))
             case Right(v)    => Right(v) // No need to invoke the callback, sync resumption will take place
--- a/zio-interop-cats/shared/src/main/scala/zio/interop/package.scala
+++ b/zio-interop-cats/shared/src/main/scala/zio/interop/package.scala
@@ -34,21 +34,25 @@ package object interop {
   @inline def toEffect[F[_], R, A](rio: RIO[R, A])(implicit R: Runtime[R], F: Async[F], trace: Trace): F[A] =
     F.defer {
       val interrupted = new AtomicBoolean(true)
-      F.async[Exit[Throwable, A]] { cb =>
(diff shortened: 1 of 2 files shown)


zio-interop-cats/shared/src/main/scala/zio/interop/package.scala
```diff
@@ -34,21 +34,35 @@ package object interop {
...
+          val out = R.unsafe.runOrFork {
             signalOnNoExternalInterrupt {
               rio
             }(ZIO.succeed(interrupted.set(false)))
           }
-          fiber.unsafe
-            .addObserver(exit => cb(Right(exit)))
-          val cancelerEffect = F.delay {
-            val _ = fiber.interrupt
+          out match {
```
steinybot, line 46: Cats syntax is imported so you could use `out.leftMap`

zio-interop-cats/shared/src/main/scala/zio/interop/package.scala
```diff
@@ -34,21 +34,35 @@ package object interop {
...
+            case Left(fiber) =>
+              val completeCb = (exit: Exit[Throwable, A]) => cb(Right(exit))
+              fiber.unsafe.addObserver(completeCb)
+              Left(Some(F.async[Unit] { cb =>
```
steinybot, line 50: Does the finalizer need to wait for the fiber complete?

zio-interop-cats/shared/src/main/scala/zio/interop/package.scala
```diff
@@ -34,21 +34,35 @@ package object interop {
...
+                F.delay {
+                  val interruptCb = (_: Exit[Throwable, A]) => cb(Right(()))
+                  fiber.unsafe.addObserver(interruptCb)
+                  fiber.unsafe.removeObserver(completeCb)
+                  fiber.tellInterrupt(Cause.interrupt(fiber.id))
+                  // Allow the interruption to be interrupted
```
steinybot, line 56: It looks like finalizers are uncancellable so this will never happen.

steinybot (commented):
Nice! Thanks for looking into this. A few passing comments/questions.

zio-interop-cats/shared/src/main/scala/zio/interop/package.scala
kyri-petrou: I thought about it but I decided against it as it's going to increase allocations from the Function1 and the closures of local vals captured within the function. I think we can afford a bit of "uglier code" to improve performance, wdut?

zio-interop-cats/shared/src/main/scala/zio/interop/package.scala
kyri-petrou: I don't think it does. The finalizer is wrapped in `delay` so it's only invoked when cancellation is invoked. Is that explaining what you asked or did I miss something?

zio-interop-cats/shared/src/main/scala/zio/interop/package.scala
kyri-petrou: Huh thanks for pointing this out. I was debating with my own self whether to add it 😬

Only reason I added it was _just in case_ CE adds support for timing out interruption. I'm happy to remove it though to make the code more readable. Wdut?

zio-interop-cats/shared/src/main/scala/zio/interop/package.scala
steinybot: My understanding is that this finalizer is `F.async` because we want it to complete only when the ZIO fiber completes. I was wondering if it could just be `F.defer` and be more of a fire and forget as in:
```
              Left(Some(F.delay {
                fiber.tellInterrupt(Cause.interrupt(fiber.id))
              }))
```

zio-interop-cats/shared/src/main/scala/zio/interop/package.scala
steinybot: I'm leaning towards removing it, especially if we don't need the finalizer to be async. It also seems unlikely that they would ever add it because it would turtles all the way down. Could you then cancel the cancellation of the cancellation?

kyri-petrou: > Could you then cancel the cancellation of the cancellation

My thinking is that some frameworks allow you to set a deadline for cancellation. But yeah I think let's remove it since it's a bit of an overkill for something that's not supported yet

kyri-petrou:
Closing and reopening as an attempt for the `/claim` tag to work

kyri-petrou:
@steinybot closing and reopening didn't seem to work, I'll close it, address the comments and open it as a new PR

zio-interop-cats/shared/src/main/scala/zio/interop/package.scala
steinybot: How does making it async prevent resource leakage? The only way for something to leak is if the ZIO fiber ignores the interupt, which is an entirely different problem. And if that does ever happen by waiting this is now leaking a cats fiber so I think it makes it worse not better.

Also if the CIO is used in a race but the ZIO takes a while to respond to the interrupt then it is holding up the downstream effects unnecessarily. Actually it might be worth seeing whether CIO race waits or not.

As for the test, can't it add an onExit to the ZIO that completes a promise and wait for that? That's what I started doing in my repro.

kyri-petrou: The async part is so that we don't synchronously block till interruption is completed.

Let's forget about interop for a minute. If we were working purely with ZIO or CIO, wouldn't you expect that all finalisers have finished executing when you issue an interrupt signal and await for the fiber to complete? This is what the async part does in a nutshell; it makes sure that everything in the interrupted effect has finalised by the time we resume execution.

steinybot: It's a good question. I can think of cases where I would want both. Looks like `IO.race` vs `IO.racePair`. The former will cancel looser and await its finalisation, the later will not. Similarly with `ZIO.race`. So yeah you are correct.

PR #760, opened 2026-02-28
ISSUE-675: Scala Native Cross-building

### LINKED ISSUE #675: Add ScalaNative crossbuilding
The cats libraries support Scala Native, and ZIO core does too, it would make sense to support ScalaNative as a platform as well.

davesmith00000:
Attempt to resolve https://github.com/zio/interop-cats/issues/675, as I require it for some work I'm doing.

The effort so far really just does the grunt work of setting up the Native builds.

There are some problems though, and I'm not really familiar enough with how all the moving parts are going to behave to be sure on the solutions. I'll try and look into it but assistance is very welcome.

We're got two broken test suites (`CatsSpec` and `ZStreamSpec`), and I had to put in a fake interface to account for an `fs2` feature (`UnixSockets.fromAsync`) that is apparently missing for Scala Native, but is possibly deprecated anyway in favour of `Network`. 🤷

You'll also notice the use of some pre-release fs2 and CE3 versions with the Scala Native builds. I confirmed with the Typelevel folks that both versions are solid and stable, they just haven't hit the release button yet.

My hunch on the broken test suites is that the native stack probably behaves differently to the JVM / JS versions. The question (assuming I'm right), is what is a reasonable amendment to account for that.

I have published this version locally and tried it on my project, and it seems to work in my limited use-case. So that's promising..


davesmith00000:
I'll leave this here in case it holds any value, but feel free to close / reject. I think I've changed my mind about needing it.

davesmith00000:
Hmmm. The build passed. Unexpected. 🤔

neko-kai:
@davesmith00000  I'm definitely not against supporting Scala Native. This'll need some investigation wrt tests seemingly working on CI that don't work for you locally, but if it builds we'll ship it. 👍

davesmith00000:
Well.. if you ship it, I'll use it - if that's any incentive at all. 😄

build.sbt
```diff
@@ -59,7 +65,11 @@ val disciplineScalaTestVersion = "2.3.0"
 val fs2Version                 = "3.12.2"
 val scalaJavaTimeVersion       = "2.6.0"
 
-lazy val zioInteropTracer    = crossProject(JSPlatform, JVMPlatform)
+// Native currently requires pre-release versions of fs2 and CE3
+val catsEffectNativeVersion = "3.7.0-RC1"
+val fs2NativeVersion        = "3.13.0-M8"
```
davesmith00000, line 70: New versions of both these libs have now been officially released! ...but I haven't tested them out on my own work yet. I'll try and think on and bump them when I have more time. 🙂

